PR #59, opened 2020-04-30
[FEATURE] Set server cmd update

GuillaumeFalourd:
**- What I did**

- Added the server URL health validation on login and set server commands (Expecting HttpStatus 200).

**- How I did it**

- Implemented a new check using the http.get() method

**- How to verify it**

- With Unit Tests

- Addition of the server URL health validation

### DIFF
--- a/pkg/security/secteam/login.go
+++ b/pkg/security/secteam/login.go
@@ -1,6 +1,7 @@
 package secteam
 
 import (
+	"fmt"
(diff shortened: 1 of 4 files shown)


GuillaumeFalourd:
@viniciusramosdefaria, just to let you know that we need the server URL to return HttpStatus 200 to approve and validate this feature.

PR #67, opened 2020-05-05
Feature/warning

victor-schumacher:
**- What I did**
 Added a message when a repository is overwritten 
**- How I did it**
When add a repository, checks if already exists 
**- How to verify it**
rit add repo <reponame>
the added repository must already exist

### DIFF
--- a/pkg/cmd/add_repo.go
+++ b/pkg/cmd/add_repo.go
@@ -65,6 +65,9 @@ func (a addRepoCmd) runFunc() CommandRunnerFunc {
 		}
 
 		list, err := a.List()
+		if err != nil {
+			return err
+		}
 		for _, repo := range list {
 			if rn == repo.Name {
 				fmt.Printf("Your repository %s has been overwritten.\n", repo.Name)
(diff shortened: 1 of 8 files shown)


victor-schumacher:
dale

pkg/cmd/add_repo.go
```diff
@@ -61,6 +67,22 @@ func (a addRepoCmd) runFunc() CommandRunnerFunc {
...
+			return err
+		}
+		for _, repo := range list {
+			if rn == repo.Name {
+				fmt.Printf("Your repository %s is gonna be overwritten.\n", repo.Name)
+				options := []string{"yes", "no"}
+				choice, _ := a.Bool("Want to proceed?", options)
+				if !choice {
+					fmt.Println("Operation cancelled")
+					return err
```
kaduartur, line 81: you could return nil because it isn't an error

pkg/cmd/add_repo.go
```diff
@@ -61,6 +67,22 @@ func (a addRepoCmd) runFunc() CommandRunnerFunc {
 			TreePath: ur,
 		}
 
+		list, err := a.List()
```
kaduartur, line 70: could you rename list to repos or repositories?

PR #90, opened 2020-05-14
Feature/dockerfile create f

victor-schumacher:
**- What I did**
Added Dockerfile templates when creating formulas
**- How I did it**
Creating a createDockerfile func 
**- How to verify it**
use ```rit create formula``` and check if the Dockerfile is there


victor-schumacher:
[Rit create formula with Docker](https://trello.com/c/SqYs3QTK/390-rit-create-formula-with-docker)

PR #102, opened 2020-05-20
Feature/functional test team

dmbarra:
**- What I did**
Created a suite of functional tests to validate core functionalities for TEAM mode.
Changed pipeline workflow to run tests after build.
Change the make to accept single and team functional tests

**- How I did it**
For the functional_test I create two separate folders fo have the single and team suite. For now, we just have a small number of scenarios for the team, because we need to change the way we do the login.

**- How to verify it**
Make functional-test-single
Make functional-test-team
Follow the pipeline build


functional/runner.go
```diff
@@ -21,8 +21,8 @@ type Step struct {
 }
 
 type Scenario struct {
-	Entry string `json:"entry"`
-	Steps []Step `json:"steps"`
+	Entry string  `json:"entry"`
+	Steps []Step  `json:"steps"`
 	Result string `json:"result"`
 }
```
victor-schumacher, line 27: i think we get some extra spaces here

functional/single/core_integration_test.go
```diff
@@ -1,16 +1,17 @@
-package functional
+package single
 
 import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
+	"github.com/ZupIT/ritchie-cli/functional"
 )
```
victor-schumacher, line 8: a empty line between 6 and 7 would be nice

PR #123, opened 2020-05-27
[FEATURE] cmd init

sandokandias:
**- What I did**
Add a new command`init` that will run a initial setup for `rit`.
 - Single: Requires the passphrase
 - Team: Requires the name and serverURL of the organization. It will possible to peform login too (optional).

**- How I did it**
 - Add a new cobra command `init` supporting prompt inputs and stdin
 - Change PreRun func in cobra command `root`
 - Change `serverURL string` to `server.Config struct` with fields _(Organization, URL)_
 - Remove unecessary files

**- How to verify it**
 - clean the rit home
 - run any command, check the output message
 - run `rit init` and perform the setup 

A new command`init` that will run a initial setup for `rit`.


pkg/cmd/init.go
```diff
@@ -0,0 +1,199 @@
...
+
+		if err := o.Set(cfg); err != nil {
+			return err
+		}
+
+		y, err := o.Bool(MsgLogin, []string{"no", "yes"})
+		if err != nil {
+			return err
+		} else if y {
+			if err := o.Login(); err != nil {
```
victor-schumacher, line 171: Make sense use else after a if return statment?

PR #124, opened 2020-05-28
Feature/improvements circleci

ernelio:
**- What I did**

 - Improvements in circleci

**- How to verify it**

- run circleci

- Improvements in circleci

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,10 +64,10 @@ jobs:
           command: |
             download=https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh
             wget -O- -q $download | sh -s -- -b /go/bin/ v1.24.0
-      - run: go mod download
       - run:
           name: lint
           command: |
+            go mod download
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -116,6 +116,7 @@ jobs:
       - run:
           name: Running Functional tests for single
           command: |
+            go mod download
```
viniciusramosdefaria, line 119: Why this step was added to everything? I don't understand

ernelio: This is a way to optimize the download of project dependencies.

.circleci/config.yml
```diff
@@ -198,12 +201,13 @@ workflows:
       - unit_test:
           filters:
             <<: *FILTERS_CHECK
+          requires:
+            - lint
```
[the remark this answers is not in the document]
ernelio: Researching the failure that was occurring with lint, it was verified that it may be due to lack of memory, so I removed the parallelism between build and test_unit.

PR #126, opened 2020-05-28
Feature/stdin functional tests

dmbarra:
**- What I did**
Created a suite of functional tests to validate STDIN commands for Single version

**- How I did it**
For the functional_test I created the same structure I did before for another test, I create Json to have scenarios and create a file to run this scenario

**- How to verify it**
Make functional-test-single

### DIFF
--- a/functional/runner.go
+++ b/functional/runner.go
@@ -88,30 +88,38 @@ func (scenario *Scenario) RunSteps() (string, error) {
 func (scenario *Scenario) RunStdin() (string, error) {
 	fmt.Println("Running: "+ scenario.Entry)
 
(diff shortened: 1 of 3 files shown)


functional/runner.go
```diff
@@ -107,15 +133,32 @@ func FuncValidateLoginRequired() {
...
+		return nil, err, nil, cmdIn
+	}
+
+	stdout, _ := cmdIn.StdoutPipe()
+
+	err = cmdIn.Start()
+	if err != nil {
+		return nil, err, nil, cmdIn
+
+	}
```
victor-schumacher, line 157: Is possible to use the sintax
```
if err = cmdIn.Start();err != nil {
		return nil, err, nil, cmdIn
	}
```
here?

PR #165, opened 2020-06-16
Feature/rit upgrade

viniciussousazup:
**- What I did**
create a upgrade command to download the binaries of last stable version, and add a warning to check if current version is equals to stable version

**- How I did it**
add rit upgrade
add PostRunFunc to check stable version

**- How to verify it**
 run upgrade
 if your version is different of stable version should print a warning

rit now have a command to upgrade the binaries to last stable version (rit upgrade) and if your version is not the last stable version a warning is printed after every command.


pkg/cmd/upgrade.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/ZupIT/ritchie-cli/pkg/api"
+	"github.com/ZupIT/ritchie-cli/pkg/prompt"
+	"github.com/ZupIT/ritchie-cli/pkg/version/versionutil"
+)
+
+const (
+	upgradeUrlFormat = "https://commons-repo.ritchiecli.io/%s/%s/%s/rit"
+)
+
+type UpgradeUtil interface {
```
marcosgmgm, line 22: separate interface in other file

pkg/cmd/upgrade.go
```diff
@@ -0,0 +1,94 @@
...
+func (u DefaultUpgradeUtil) Apply(reader io.Reader, opts update.Options) error {
+	return update.Apply(reader, opts)
+}
+
+type UpgradeCmd struct {
+	upgradeUrl  string
+	upgradeUtil UpgradeUtil
+}
+
+func GetUpgradeUrl(edition api.Edition, resolver versionutil.Resolver) string {
```
marcosgmgm, line 37: Remove Get of func name

pkg/file/fileutil/file_util_service.go
```diff
@@ -0,0 +1,16 @@
+package fileutil
```
marcosgmgm, line 1: use pkg/file/fileutil/file_util.go

pkg/version/versionutil/version.go
```diff
@@ -0,0 +1,103 @@
+package versionutil
```
marcosgmgm, line 1: change folder to pkg/version

pkg/version/versionutil/version.go
```diff
@@ -0,0 +1,103 @@
...
+	FileUtilService  fileutil.FileUtilService
+	HttpClient       *http.Client
+}
+
+type stableVersionCache struct {
+	StableVersion string `json:"stableVersion"`
+	ExpiresAt     int64  `json:"expiresAt"`
+}
+
+func (r DefaultVersionResolver) GetCurrentVersion() (string, error) {
```
marcosgmgm, line 41: Remove Get of func name and separated interface in other file

viniciussousazup: done : interface separated to other file
to do : Remove Get of func name

pkg/version/versionutil/version.go
viniciussousazup: changed to version_util

pkg/cmd/upgrade.go
viniciussousazup: create file upgrade on pkg/upgrade

PR #190, opened 2020-06-22
[FEATURE] Change login and init to get totp

viniciussousazup:
**- What I did**
when run rit init save server otp value
when run login get otp value and ask if needed

**- How I did it**
change init cmd
change login cmd

**- How to verify it**
see server.json and verrify otp value
try to login on a server with otp

add otp support on cli


viniciussousazup:
waiting server new version to merge

PR #212, opened 2020-06-25
[FEATURE] Create formula and build

kaduartur:
**- What I did**
- Add the new template for build formula inside Windows named build.bat
- Create packages inside the formula package to organize it
- Add omitempty to Command and Formula struct
- Add validation when RepoUrl does not exist in tree.json 

**- How to verify it**
Run tests and run the command `rit create formula` to check it


pkg/formula/repo.go
```diff
@@ -12,36 +12,36 @@ type RepositoryFile struct {
...
 
-type AddLister interface {
-	Adder
-	Lister
+type RepoAddLister interface {
+	RepoAdder
+	RepoLister
 }
 
 type DelLister interface {
```
sandokandias, line 44: RepoDelLister?

Issue #173, opened 2020-06-17
[BUG] Error when adding kubeconfig too big


erneliojuniorzup:
**What happened**:

**What you expected to happen**:

`Kubeconfig credential saved!`

**How to reproduce it (as minimally and precisely as possible)**:

`rit set credential`
`Provider: kubeconfig`
`credential: base64config="kubeconfig in base64 a long time"` 

**Environment**:

- Ritchie version (use `rit --version`):

`rit version 1.0.0-beta.14 (single)
  Build date: 06/15/20_14:00
  Built with: go1.14.4
`

- Operating System:

`NAME="Ubuntu"
VERSION="18.04.4 LTS (Bionic Beaver)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 18.04.4 LTS"
VERSION_ID="18.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=bionic
UBUNTU_CODENAME=bionic`

victor-schumacher:
![](https://github.trello.services/images/mini-trello-icon.png) [[FIX] Set credential too long text](https://trello.com/c/lgQvueCX/481-fix-set-credential-too-long-text)

viniciusramosdefaria:
Closing issue, code already in review.

PR #221, opened 2020-06-29
[FEATURE] Added file entry to credentials

brunats:
**- What I did**
I added the entry via file in the credentials, now it is possible to add the value of one credential per .txt file instead of the command line.
https://github.com/ZupIT/ritchie-cli/issues/173

**- How I did it**
Adding an extra step after defining the provider, the system now asks if the credential entry will be per terminal or per file.

The case of being by terminal remains the same old operation.
If it is a file, the system asks for the credential key and then the path to the file that contains its value.

The system then reads the value contained in the file and proceeds to normal validations (if it is not empty).

**- How to verify it**
Commands:
```Shell
rit set credential
```

Add any test provider, choose input via file - "File (input by file)" and proceed with the system steps.

teste-pr-entry-file

Added file entry to credentials


pkg/cmd/set_credential_test.go
```diff
@@ -27,3 +28,36 @@ func TestNewTeamSetCredentialCmd(t *testing.T) {
...
+	)
+	cmd.PersistentFlags().Bool("stdin", false, "input by stdin")
+	if cmd == nil {
+		t.Errorf("NewTeamSetCredentialCmd got %v", cmd)
+	}
+
+	if err := cmd.Execute(); err != errPath {
+		t.Errorf("%s = '%v', want '%v'", cmd.Use, err, errPath)
+	}
+}
```
victor-schumacher, line 63: It is possible to use %q instead of '%v' here?

brunats: Is possible, but I don't know if it's necessary.

An example with `%q`:
```Go
	if err := cmd.Execute(); err != another {
		t.Errorf("%s = '%q', want '%q'", cmd.Use, err, another)
	}
```
```Shell
--- FAIL: TestNewSingleSetCredentialCmdWithEntryArchive (0.00s)
    set_credential_test.go:61: credential = '"another error"', want '"some error of path"'
```

And with  `%v`:

```Shell
--- FAIL: TestNewSingleSetCredentialCmdWithEntryArchive (0.00s)
    set_credential_test.go:61: credential = 'another error', want 'some error of path'
```


What would be the motivation you thought for the exchange?

victor-schumacher: The idea is to remove the ' ' and use only %q, since you are quoting something.

brunats: Yes makes sense, I'll make the change.

brunats: Fix in https://github.com/ZupIT/ritchie-cli/pull/221/commits/3be8e6adbfbd8ec7a73fec3b9da44f6319069917

pkg/cmd/set_credential.go
```diff
@@ -145,6 +155,62 @@ func validate(pair []string) string {
 	return ""
 }
 
+func (s setCredentialCmd) entryCredential() ([]string, error) {
+	var pair []string
+	var err error
+	var kv string
+
+	entries := map[string]string{
+		EntriesTypeCredentialFile:   "file",
```
sandokandias, line 164: extract a const for string `"file"`

pkg/cmd/set_credential.go
```diff
@@ -145,6 +155,62 @@ func validate(pair []string) string {
...
+	for k := range entries {
+		types = append(types, k)
+	}
+
+	typ, errTyp := s.List(MsgTypeEntry, types)
+	if errTyp != nil {
+		return nil, errTyp
+	}
+
+	if entries[typ] == "file" {
```
sandokandias, line 178: const

pkg/cmd/set_credential.go
```diff
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"io/ioutil"
```
sandokandias, line 5: You should use the interface `FileReader` in https://github.com/ZupIT/ritchie-cli/blob/master/pkg/stream/file.go and pass as dependency

pkg/cmd/set_credential.go
```diff
@@ -26,6 +27,16 @@ type setCredentialCmd struct {
 	prompt.InputMultiline
 }
 
+// MsgTypeEntry used in select of type entry credential
```
sandokandias, line 30: The comment must be above of the const name

pkg/cmd/set_credential_test.go
```diff
@@ -1,7 +1,13 @@
 package cmd
 
 import (
+	"errors"
+	"io/ioutil"
```
sandokandias, line 5: You should use the interface FileReader in https://github.com/ZupIT/ritchie-cli/blob/master/pkg/stream/file.go and create a mock for it

brunats: Fix in https://github.com/ZupIT/ritchie-cli/pull/221/commits/87aeba9dcd93be746056c65f2e8f41f6186d6e21

pkg/cmd/set_credential.go
brunats: Fix in https://github.com/ZupIT/ritchie-cli/pull/221/commits/87aeba9dcd93be746056c65f2e8f41f6186d6e21

pkg/cmd/set_credential.go
brunats: Fix in https://github.com/ZupIT/ritchie-cli/pull/221/commits/a1563cf4c193bd57463f548675d6079bb4c955ae, I decided to remove the comment, because the right thing would be to include a comment for each constant and I saw that it would be polluted without really adding.

pkg/cmd/set_credential.go
brunats: Fix in https://github.com/ZupIT/ritchie-cli/pull/221/commits/a1563cf4c193bd57463f548675d6079bb4c955ae

pkg/cmd/set_credential.go
brunats: Fix in https://github.com/ZupIT/ritchie-cli/pull/221/commits/a1563cf4c193bd57463f548675d6079bb4c955ae

sandokandias:
closed in favor of [287](https://github.com/ZupIT/ritchie-cli/pull/287)

PR #222, opened 2020-06-29
[Feature] Feature/output

viniciussousazup:
**- What I did**
Created RIT_OUTPUT_DIR and passed this env to the formulas, with this env the formula can create files, thats files is called as outputs then this outputs can be logged,printed or used on a workflow

**- How I did it**
create output env
create output dir on tmp bin dir
ignore output dir when copy tmp bin dir to user workdir
change config.json to add outputs fields

**- How to verify it**
create some formula and use outputs fields

add formula outputs


pkg/formula/runner/outputs.go
```diff
@@ -0,0 +1,66 @@
...
+func NewOutputManager(
+	w io.Writer,
+) OutputManager {
+	return OutputManager{
+		writer: w,
+	}
+}
+
+func (o OutputManager) ValidAndPrint(setup formula.Setup) error {
+	_, err := o.writer.Write([]byte(printAndValidOutputDir(setup)))
```
kaduartur, line 26: Could you call the printAndValidOutputDir function before Write to make it clean?

pkg/formula/runner/outputs.go
```diff
@@ -0,0 +1,66 @@
...
+	return err
+}
+
+func printAndValidOutputDir(setup formula.Setup) string {
+
+	files, err := ioutil.ReadDir(setup.TmpOutputDir)
+	if err != nil {
+		return prompt.Red("Fail to read output dir")
```
kaduartur, line 34: Could you create exportable error and return (string, error) in printAndValidOutputDir?

pkg/formula/runner/outputs.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	fOutputs := map[string]string{}
+
+	resolveKey := func(name string) string { return strings.ToUpper(name) }
+
+	if len(files) != len(setup.Config.Outputs) {
+		return prompt.Red("Output dir size is different of outputs array in config.json")
```
kaduartur, line 41: Could you create an exportable error?

pkg/formula/runner/post_run.go
```diff
@@ -40,6 +44,10 @@ func (PostRunnerManager) PostRun(p formula.Setup, docker bool) error {
 	return nil
 }
 
+func RemoveTmpOutputDir(p formula.Setup) error {
```
kaduartur, line 47: do you need this function to be public?

PR #232, opened 2020-07-01
[FEATURE] Build formulas on Windows

kaduartur:
**- What I did**

I create scripts to build formulas on Windows in the following languages:

- Java
- Node
- Php
- Python

**- How I did it**

I created templates to build.bat file, inside the `pkg/formula/creator/templates` for all supported languages.

**- How to verify it**
On Windows, run the command `rit create formula`. You can run the `rit build formula` command to get the same result too. 
After that, you can run your formula on windows

Create scripts to build formulas on Windows


pkg/formula/creator/lang.go
```diff
@@ -10,9 +10,9 @@ import (
 	"github.com/ZupIT/ritchie-cli/pkg/formula/creator/templates/template_go"
```
sandokandias, line 10: I think that we should to avoid using `_` in package names.

Source: https://golang.org/doc/effective_go.html#package-names

pkg/formula/creator/creator.go
```diff
@@ -18,7 +18,7 @@ import (
 
 	"github.com/ZupIT/ritchie-cli/pkg/api"
 	"github.com/ZupIT/ritchie-cli/pkg/file/fileutil"
-	"github.com/ZupIT/ritchie-cli/pkg/formula/creator/templates/template_go"
+	"github.com/ZupIT/ritchie-cli/pkg/formula/creator/templates/golang"
```
sandokandias, line 21: what if we follow the golang std?

`golang/template`

e.g: https://golang.org/pkg/html/template/

PR #250, opened 2020-07-07
Feature/verbose flag

antonioolf:
**- What I did**

Added the possibility to execute a formula passing a `--verbose` flag to indicate that formula should show more detailed log messages. 

This verbose flag only adds a Boolean environment variable that can be retrieved during the execution of the formula, so the details of the messages are in the domain of whoever is writing the formula.

**- How to verify it**

- Run a formula with `--verbose` as flag
- In the formula script, retrieve the environment variable `VERBOSE_MODE`
- If the flag has been entered, the value will be `1`, otherwise it will be `0` (default)

Example:
`rit docker generate compose --verbose`

Adding `--verbose` flag


antonioolf:
Working on code coverage now

Issue #186, opened 2020-06-22
[FEATURE] Adding hello world template in RUBY (rit create formula)


GuillaumeFalourd:
## What would you like to be added:

A Hello World template to create a formula in RUBY with the rit create formula command

## Why is this needed:

Those templates are currently available for Java, Go, Python, Node and Shell. It would make it easier for RUBY programmers to create formulas and use Ritchie.

PR #252, opened 2020-07-08
[Feature] Adding ruby language support

henriquemoraes8:
**- What I did**
Added ruby language support for when creating a new formula according to issue #186 . **I have intentionally left out the coffee scaffold as I think it will be discontinued**, if not, please let me know

To provide more coverage on fail cases, I had to resort to dependency injection on `createSrcFiles` by extracting an interface on `CreateManager`. Ended up running this fail case test to all other languages and raised the coverage on `createSrcFiles` from 75% to 92.9%

Removed an extraneous param used on the languages `New` function

Fixed a small bug in which failed `Shell` language creation would return `nil` and not account for the error

**- How to verify it**
you can run `rit create formula` and then select the `ruby` language. The appropriate `rb` files along with the `Gemfile` are created. You can also verify it through `make unit-test`

Screen Shot 2020-07-07 at 15 52 11

* Added support to write new formulas in ruby


pkg/formula/creator/lang/golang/golang.go
```diff
@@ -17,12 +17,10 @@ type Go struct {
 }
 
 func New(
-	c formula.Creator,
```
henriquemoraes8, line 20: These dont seem to be necessary, the creator interface requires `Create` method to be implemented and each file implements its own already, so it is not composing from the passed argument

pkg/formula/creator/lang/ruby/ruby.go
```diff
@@ -0,0 +1,70 @@
...
+		return err
+	}
+
+	if err := createGemfile(srcDir, n.Gemfile); err != nil {
+		return err
+	}
+
+	templateNode := strings.ReplaceAll(n.File, formula.NameBin, pkg)
+	pkgFile := fmt.Sprintf("%s/%s%s", pkgDir, pkg, n.FileFormat)
+	if err := fileutil.WriteFile(pkgFile, []byte(templateNode)); err != nil {
```
victor-schumacher, line 61: This variable should be templateRuby, right?

henriquemoraes8: you are right, thanks!

PR #256, opened 2020-07-08
FEATURE - Improving information about help

brunats:
**- What I did**
Updated the help messages for the core commands and added comments to some public functions found in the modified files.

**- How I did it**

Changing the help messages according to the documentation and adding comments on the necessary public functions (in files already changed previously).

**- How to verify it**

Exec `rit`
Before:
antes

Now:
depois

> Prints updated on Jul 14 at 7:03 pm

Improving information about help


brunats:
> Good job! It became much clearer with examples.
> Do you think it is a good idea not to use the word "objects"? For me it is not valuable information.


Looking again, I agree with you, let's see what the other reviewers think and depending on I remove the 'objects'.

pkg/cmd/autocomplete.go
```diff
@@ -22,7 +22,7 @@ type autocompleteCmd struct {
 func NewAutocompleteCmd() *cobra.Command {
 	return &cobra.Command{
 		Use:     "completion SUBCOMMAND",
-		Short:   "Add autocomplete for terminal",
+		Short:   "Add autocomplete for terminal (bash, zsh)",
```
marcoscostazup, line 25: We have autocomplete for fish and powershell too now

brunats: I am updating to add them through a list.

brunats: Fix in https://github.com/ZupIT/ritchie-cli/pull/256/commits/6c3bf279b21ee10801679832104b3d8aa3364d33.

marcoscostazup: Great! :)

PR #268, opened 2020-07-10
[FEATURE] Added `rit delete repo` command and added to list repo a total info a…

marcoscostazup:
…t the end

**- What I did**

**- How to verify it**


pkg/cmd/delete_repo.go
```diff
@@ -0,0 +1,60 @@
...
+
+func (dr DeleteRepoCmd) runFunc() CommandRunnerFunc {
+	return func(cmd *cobra.Command, args []string) error {
+		repos, err := dr.RepositoryLister.List()
+		if err != nil {
+			return err
+		}
+
+		if len(repos) == 0 {
+			prompt.Warning("You should add a repository first")
```
kaduartur, line 41: Edit the message to `You don't have any repositories`

pkg/cmd/delete_repo.go
```diff
@@ -0,0 +1,60 @@
...
+			return nil
+		}
+
+		var reposNames []string
+		for _, r := range repos {
+			reposNames = append(reposNames, r.Name)
+		}
+
+		repo, err := dr.InputList.List("Repository:", reposNames)
```
kaduartur, line 50: Check InputList err

pkg/cmd/delete_repo.go
```diff
@@ -0,0 +1,60 @@
...
+
+		err = dr.Delete(repo)
```
kaduartur, line 52: The good practice is `if err := dr.Delete(repo); err != nil { return err }` when a function return just one error

pkg/formula/repo/deleter.go
```diff
@@ -0,0 +1,79 @@
...
+func NewDeleter(ritHome string, file stream.FileWriteReadExister, dir stream.DirRemover) DeleteManager {
+	return DeleteManager{
+		ritHome: ritHome,
+		file:    file,
+		dir:     dir,
+	}
+}
+
+func (dm DeleteManager) Delete(repoName string) error {
+	err := dm.deleteFromReposFile(repoName)
```
kaduartur, line 26: The good practice is check error in the same line when a function returns just one error.

pkg/formula/repo/deleter.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = dm.deleteRepoDir(repoName)
```
kaduartur, line 31: The good practice is check error in the same line when a function returns just one error.
Remove Dir before removing repo from repositories.json.

pkg/formula/repo/deleter.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (dm DeleteManager) deleteRepoDir(repoName string) error {
+	path := path.Join(dm.ritHome, reposDirName, repoName)
+	err := dm.dir.Remove(path)
```
kaduartur, line 40: The good practice is check error in same line when a function returns just one error.

pkg/formula/repo/lister.go
```diff
@@ -18,6 +18,7 @@ func NewLister(ritHome string, file stream.FileReadExister) ListManager {
 	return ListManager{ritHome: ritHome, file: file}
 }
 
+// Returns an empty repo if there is no repositories.json
```
kaduartur, line 21: The good practice for comments in Golang is to start the comment  with the function name

pkg/cmd/delete_repo.go
marcoscostazup: Ok! Done

pkg/cmd/delete_repo.go
marcoscostazup: Ok! Done

pkg/cmd/delete_repo.go
marcoscostazup: Ok! Done

pkg/formula/repo/deleter.go
marcoscostazup: Ok! Done

pkg/formula/repo/deleter.go
marcoscostazup: I was thinking about remove the repo from repositories.json first because if it pass ok and remove dir returns some error the user will not be able to select that dir anyway, and if he adds that repo again the directory will be overlaid.

If I remove from dir first and get some error on removing from repositories.json it would still apear when I run `rit list repo` for example (as I just look to repositories.json)

But I may be wrong, what do you think?

marcoscostazup: Actually, maybe we should return nil even if deleteRepoDir returns some error... The directory will not be erased, but if we return an error on Delete too it would not be all true, because the user would not be able to see that repo on the other commands (e.g. `rit set repo-priority` or `rit list repo`)

pkg/formula/repo/deleter.go
marcoscostazup: Ok! Done

pkg/formula/repo/lister.go
marcoscostazup: Ok! Done

PR #283, opened 2020-07-15
Add context to docker runner and -it arg to docker run

JoaoDanielRufino:
**- What I did**
Added rit context as ENV variable when running a formula with docker and -it arg to docker run to keep stdin open and allocate a pseudo-TTY

**- How to verify it**
Run any formula with docker and check the env CONTEXT

Add context to docker runner and -it arg to docker run


viniciusramosdefaria:
Please limit the scope of your modifications. Just change the things regarding docker. Other things like templates are being handled by another issue.

PR #287, opened 2020-07-16
Add files entry to `set credential` command

marcoscostazup:
**- What I did**
- Added files entry to `rit set credential` command

**- How to verify it**
- run `rit set credential`

- Added files entry to `rit set credential` command


pkg/cmd/set_credential.go
```diff
@@ -171,26 +183,54 @@ func (s setCredentialCmd) singlePrompt() (credential.Detail, error) {
...
-			if err != nil {
-				return credDetail, err
+			var value string
+			if i.Type == inputTypes[1] {
+				value, err = s.Password(i.Name + ":")
+				if err != nil {
+					return credDetail, err
+				}
+			} else {
+				value, err = s.Text(i.Name, true)
```
victor-schumacher, line 206: What do you think about using (i.Name + ":", true) on 206? So it matches with 201.

marcoscostazup: That's really better!

PR #294, opened 2020-07-17
adding powershell and fish and modifying helpers

viniciusramosdefaria:
**- What I did**
Added missing helpers and autocomplete

**- How to verify it**
See if `rit completion` reflects:
```
Add autocomplete for terminal, Available for (zsh, bash, fish, powershell).

Examples:
rit completion zsh

Available Commands:
  bash        Add bash autocomplete for terminal, --help to know how to use
  fish        Add fish autocomplete for terminal, --help to know how to use
  powershell  Add powerShell autocomplete for terminal, --help to know how to use
  zsh         Add zsh autocomplete for terminal, --help to know how to use

```

### DIFF
--- a/pkg/cmd/autocomplete.go
+++ b/pkg/cmd/autocomplete.go
@@ -40,9 +40,19 @@ func NewAutocompleteZsh(g autocomplete.Generator) *cobra.Command {
 
 	return &cobra.Command{
 		Use:     zsh.String(),
-		Short:   "Add zsh autocomplete for terminal",
-		Long:    "Add zsh autocomplete for terminal",
-		Example: "rit completion zsh",
+		Short:   "Add zsh autocomplete for terminal, --help to know how to use",
+		Long:    `
+Add bash autocomplete for terminal
+Only works if zsh auto completion is installed.
+
+To test run: 
+ $ rit completion zsh | source
+
+To install run: 
+ $ echo "[[ -r "/usr/local/bin/rit" ]] && rit completion zsh > ~/.rit_completion" >> ~/.zshrc
+
+`,
+		Example: "rit completion zsh | source",
 		RunE:    a.runFunc(),
 	}
 }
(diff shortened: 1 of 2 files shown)


pkg/cmd/autocomplete.go
```diff
@@ -40,9 +40,19 @@ func NewAutocompleteZsh(g autocomplete.Generator) *cobra.Command {
...
+		Short:   "Add zsh autocomplete for terminal, --help to know how to use",
+		Long:    `
+Add bash autocomplete for terminal
+Only works if zsh auto completion is installed.
+
+To test run: 
+ $ rit completion zsh | source
+
+To install run: 
+ $ echo "[[ -r "/usr/local/bin/rit" ]] && rit completion zsh > ~/.rit_completion" >> ~/.zshrc
```
viniciussousazup, line 52:
```suggestion
 $ echo "[[ -r "/usr/local/bin/rit" ]] && rit completion zsh | source" >> ~/.zshrc
```

i guess only save completion on ~/.rit_completion don't work if the user not call "source ~/.rit_completion  "

pkg/cmd/autocomplete.go
```diff
@@ -53,9 +63,19 @@ func NewAutocompleteBash(g autocomplete.Generator) *cobra.Command {
...
+		Short:   "Add bash autocomplete for terminal, --help to know how to use",
+		Long:    `
+Add bash autocomplete for terminal
+Only works if bash auto completion is installed.
+
+To test run: 
+ $ rit completion bash | source
+
+To install run: 
+ $ echo "[[ -r "/usr/local/bin/rit" ]] && rit completion bash > ~/.rit_completion" >> ~/.bashrc
```
viniciussousazup, line 75:
```suggestion
 $ echo "[[ -r "/usr/local/bin/rit" ]] && rit completion bash | source" >> ~/.bashrc
```
i guess only save completion on ~/.rit_completion don't work if the user not call "source ~/.rit_completion  "

pkg/cmd/autocomplete.go
```diff
@@ -40,9 +40,19 @@ func NewAutocompleteZsh(g autocomplete.Generator) *cobra.Command {
 
 	return &cobra.Command{
 		Use:     zsh.String(),
-		Short:   "Add zsh autocomplete for terminal",
-		Long:    "Add zsh autocomplete for terminal",
-		Example: "rit completion zsh",
+		Short:   "Add zsh autocomplete for terminal, --help to know how to use",
+		Long:    `
+Add bash autocomplete for terminal
```
viniciussousazup, line 45:
```suggestion
Add zsh autocomplete for terminal
```

pkg/cmd/autocomplete.go
viniciusramosdefaria: I added the line:

` echo "source ~/.rit_completion" >> ~/.bashrc`
To solve the missing stuff.
The reason we need to do this way is that when running on macOS stuff breaks if you do like you mentioned. I don't know why LOL.

pkg/cmd/autocomplete.go
viniciusramosdefaria: I added the line:

` echo "source ~/.rit_completion" >> ~/.zsh `

To solve the missing stuff.
The reason we need to do this way is that when running on macOS stuff breaks if you do like you mentioned. I don't know why LOL.

PR #323, opened 2020-07-30
Fix/list credential

victor-schumacher:
**- What I did**
We change how the values ​​of the credentials are shown;
Before, long credentials, in the worst case scenario, had 50% of them cut off, causing the wrong impression;
Now, we show the first 4 characters and hide the rest

**- How to verify it**
rit list credential;
You need a credential with more then 20 characters

rit list credential improvement

### DIFF
--- a/pkg/credential/settings.go
+++ b/pkg/credential/settings.go
@@ -84,7 +84,6 @@ func formatCredValue(credential string) string {
 				break
 			}
 			resumedCredential = append(resumedCredential, r)
-
 		}
 		return string(resumedCredential) + "..."
 	} else {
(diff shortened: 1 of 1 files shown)


pkg/credential/settings.go
```diff
@@ -73,15 +73,17 @@ func formatCredential(credential string) string {
 
 	return strings.Join(splitedCredential, "\"")
 }
-
```
brunasilvazup, line 76: I think this line stays, right?

victor-schumacher: @brunasilvazup 
Right. As it was already merged, i will add the line back on my next pull request. Thanks!

PR #327, opened 2020-07-30
Adding ansible credentials and bugfix for 2.0

henriquemoraes8:
- What I did

Adoption has been working on formulas for the API Manager team. One of them requires the provisioning of some hcl files to the cluster, which is done with Ansible. So it is being added to the pool of credentials

There was a bug in when adding a new credential provider, it would not be listed because the default credentials were only added if there was no config file. Upon further investigation, I think this file is not rewritten on updates as well. So the solution was to check a diff of the default credentials vs the current config.json file

- How to verify it
Run rit set credential and check for the existence of ansible option

- Description for the changelog
Added ansible support to rit set credential


pkg/credential/settings.go
```diff
@@ -130,11 +131,39 @@ func (s Settings) WriteCredentialsFields(fields Fields, path string) error {
 // WriteDefault is a non override version of WriteCredentialsFields
 // used to create providers.json if user dont have it
 func (s Settings) WriteDefaultCredentialsFields(path string) error {
-	if !s.file.Exists(path) {
-		err := s.WriteCredentialsFields(NewDefaultCredentials(), path)
-		return err
+	fieldsToWrite := NewDefaultCredentials()
+	if s.file.Exists(path) {
+		configFile, err := fileutil.ReadFile(path)
```
viniciussousazup, line 136: Deprecated: use the stream package to work with files and directories

pkg/credential/settings.go
```diff
@@ -130,11 +131,39 @@ func (s Settings) WriteCredentialsFields(fields Fields, path string) error {
...
+
+		// Avoid I/O consumption if there is nothing to change
+		if len(diff) == 0 {
+			return nil
+		}
+
+		for _, key := range diff {
+			(*credentialFields)[key] = fieldsToWrite[key]
+		}
+		fieldsToWrite = *credentialFields
```
viniciussousazup, line 163: I think this logic got a little complicated, maybe write some tests in settings_test.go ? You can make a mock of FileReader and FileWriter to assert if function  WriteDefaultCredentialsFields is right

henriquemoraes8: Hey, they are there, please check `TestSingleSettings_WriteDefaultCredentialsOnExistingFile` the test:
* Creates a temp file
* Writes a file with custom credentials
* Calls the method into this highlighted flow above
* Reads the resulting file and checks for the default credentials + custom ones
* Removes the file

I feel that using mocks we would be masking the expected behaviour here, so it is done mockless

henriquemoraes8: I have suggested separating custom from default credentials into two files and then listing them together to simplify logic. Then on every ritchie upgrade we could delete the the default and rewrite it without losing custom credentials and without using this logic. It would actually avoid this method entirely which runs on every credential call.

viniciussousazup: maybe just join the saved fields with the default fields:

```go
		// Join saved fields and default fields
		credentialFields := Fields{}
		err = json.Unmarshal(configFile, &credentialFields)
		if err != nil {
			return err
		}
		for k, v := range fieldsToWrite {
			credentialFields[k] = v
		}

		// Avoid I/O consumption if there is nothing to change
		if len(credentialFields) == len(fieldsToWrite) {
			return nil
		}

		fieldsToWrite = credentialFields
```

PR #332, opened 2020-07-30
Change build formulas behavior

kaduartur:
**- What I did**
- Removed docs dir from build formula list
- Removed `--local` and added `--docker` flag
- Check if build.bat file exist before to build a formula on Windows
- Add `--rm` in docker run command
- Check stderr when to build a formula locally on Windows
- Remove **/bin** when the formula building fail
- Now, formulas containers are built with the same formula name

**- How to verify it**
- Run unit tests
- Create and build a new formula on Windows


fabianofernandeszup:
@kaduartur Is this change to the --docker flag back to how it was in v1?
In v2 shouldn't the docker run be the default?

PR #334, opened 2020-07-31
Remove image docker post run

fabianofernandeszup:
**- What I did**
Added a command to completely remove the image from the embedded Docker after execution. It will never be reused and takes up a lot of space on the user's computer. Some with almost 2GB.
The current implementation removed only the container.

**- How to verify it**
Run any formula more than once and you will see that several builds with random ids are generated.
With this implementation, after the execution the image is removed as well as the container.

Added a complete removal of the docker image after execution, to save disk space.

Examples of garbage images before implementation:

images-docker

### DIFF
--- a/pkg/formula/runner/post_run.go
+++ b/pkg/formula/runner/post_run.go
@@ -43,6 +43,10 @@ func (po PostRunnerManager) PostRun(p formula.Setup, docker bool) error {
 		if err := removeContainer(p.ContainerId); err != nil {
 			return err
 		}
+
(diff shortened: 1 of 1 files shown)


viniciussousazup:
this pr https://github.com/ZupIT/ritchie-cli/pull/332/files already fix this bug

sandokandias:
Closed in favor of https://github.com/ZupIT/ritchie-cli/pull/332

fabianofernandeszup:
Nice.

PR #345, opened 2020-08-03
Credential through file

victor-schumacher:
**- What I did**
Added input credential througth file

**- How to verify it**
use rit set credential and choose the file option

Added input credential througth file


viniciussousazup:
set credencial tested , good job :heavy_check_mark: :rocket:

PR #350, opened 2020-08-04
Support gitlab provider on command add repo

kaduartur:
**- What I did**
- Added new provider to add repo (Gitlab)
- Separated `git` pkg in `github` and `gitlab`

**- How to verify it**
- Add new repo with `rit add repo`
- Run the unit tests with `make unit-test` command

 Add Gitlab provider to add repo

### DIFF
--- a/pkg/git/gitlab/gitlab.go
+++ b/pkg/git/gitlab/gitlab.go
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2020 ZUP IT SERVICOS EM TECNOLOGIA E INOVACAO SA
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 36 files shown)


pkg/git/gitlab/gitlab.go
```diff
@@ -0,0 +1,84 @@
...
+// TagsUrl returns the Gitlab API URL for get all tags
+// e.g. https://yourhost/api/v4/projects/{{owner}}%2F{{repo}}/releases
+func (in DefaultRepoInfo) TagsUrl() string {
+	id := url.QueryEscape(path.Join(in.owner, in.repo))
+	return fmt.Sprintf(TagsUrlPattern, in.host, id)
+}
+
+// Deprecated: Gitlab API does not implement the latest tag URL
+func (in DefaultRepoInfo) LatestTagUrl() string {
+	return ""
```
viniciussousazup, line 72: maybe peform a request with page = 1 size = 1 to list release api and get the first item of list, because List Releases return Paginated list of Releases, sorted by released_at.


https://docs.gitlab.com/ee/api/releases/

viniciussousazup: but i thing last release is used only on init that always use the commons repo

Issue #370, opened 2020-08-06
[FEATURE] Process and send metrics to server


victor-schumacher:
**What would you like to be added**:
- Receive metrics data and process them
- Create a way to anonymously identify the user on metrics.
- Send metrics to server

**Why is this needed**:

To map the behavior by the user

PR #375, opened 2020-08-07
[FEATURE] Create interface for metrics

kaduartur:
**- What I did**
> related https://github.com/ZupIT/ritchie-cli/issues/370
- Create `Sender` interface
- Create `UserIdGenerator` interface

**- How to verify it**
----
----

### DIFF
--- a/pkg/metric/metric.go
+++ b/pkg/metric/metric.go
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2020 ZUP IT SERVICOS EM TECNOLOGIA E INOVACAO SA
+ *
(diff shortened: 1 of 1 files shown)


pkg/metric/metric.go
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package metric
+
+import "time"
+
+type ID string
+
+type UserId string
```
victor-schumacher, line 22: UserID maybe?

kaduartur: we use the camel case by default, but I will fix ID on line 20.

PR #403, opened 2020-08-11
Mount volume of the .rit folder inside the container

fabianofernandeszup:
**- What I did**
I added the volume mapping from the local .rit folder to the one inside the docker container. That way the formula can use this information, such as context, credentials, core commands and even other formulas.

**- How to verify it**

Add the rit installer to a formula's docker file, run it with docker and call another local formula within the formula.

Added support for executing rit commands inside a container.


pkg/formula/runner/runner.go
```diff
@@ -92,11 +92,12 @@ func (ru RunManager) Run(def formula.Definition, inputType api.TermInputType, do
 
 func (ru RunManager) runDocker(setup formula.Setup, inputType api.TermInputType, verbose bool) (*exec.Cmd, error) {
 	volume := fmt.Sprintf("%s:/app", setup.Pwd)
+	homeDir, _ := os.UserHomeDir()
```
kaduartur, line 95: you can receive homeDir in the constructor because in the main file we have already generated it

pkg/formula/runner/runner.go
```diff
@@ -92,11 +92,12 @@ func (ru RunManager) Run(def formula.Definition, inputType api.TermInputType, do
...
 	var args []string
 	if isatty.IsTerminal(os.Stdout.Fd()) {
-		args = []string{"run", "--rm", "-it", "--env-file", envFile, "-v", volume, "--name", setup.ContainerId, setup.ContainerId}
+		args = []string{"run", "--rm", "-it", "--env-file", envFile, "-v", volume, "-v", homeDir + "/.rit:/root/.rit", "--name", setup.ContainerId, setup.ContainerId}
```
kaduartur, line 98: could you create a new variable named volumeRit and format this equal the app volume?

fabianofernandeszup:
> Could you create a new issue for this pull request?

Done.

https://github.com/ZupIT/ritchie-cli/issues/455

PR #405, opened 2020-08-11
Metrics check

victor-schumacher:
**- What I did**
Added a check function which verifies metrics status
**- How to verify it**
Check and run the unit tests
Added check function for metrics

### DIFF
--- a/pkg/cmd/metrics.go
+++ b/pkg/cmd/metrics.go
@@ -1,9 +1,11 @@
 package cmd
 
 import (
+	"path/filepath"
+
 	"github.com/spf13/cobra"
 
-	"github.com/ZupIT/ritchie-cli/pkg/metric"
+	"github.com/ZupIT/ritchie-cli/pkg/metrics"
 	"github.com/ZupIT/ritchie-cli/pkg/prompt"
 	"github.com/ZupIT/ritchie-cli/pkg/stream"
 )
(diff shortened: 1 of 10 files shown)


brunasilvazup:
> **- What I did**
> 
> **- How to verify it**
> 
> **- Description for the changelog**

Hello,

Could you add the explanatory texts for PR, please?

victor-schumacher:
> > **- What I did**
> > **- How to verify it**
> > **- Description for the changelog**
> 
> Hello,
> 
> Could you add the explanatory texts for PR, please?

Sure, I was forgetting it. Thanks.

pkg/metrics/checker.go
```diff
@@ -0,0 +1,55 @@
...
+	result := true
+	if string(bytes) == "no" {
+		result = false
+	}
+
+	return result, nil
+}
+
+func path() string {
+	hd, _ := os.UserHomeDir()
```
brunasilvazup, line 53: Here would be the `api.RitchieHomeDir ()`?

And this function is the same as 'MetricsPath', so could you leave 'MetricsPath' and use it in `Check ()`?

pkg/cmd/metrics.go
```diff
@@ -32,21 +37,22 @@ func NewMetricsCmd(file stream.FileWriteReadExister, inList prompt.InputList) *c
 
 func (m metricsCmd) run() CommandRunnerFunc {
 	return func(cmd *cobra.Command, args []string) error {
-		if !m.FileWriteReadExister.Exists(metric.MetricsPath()) {
+		path := filepath.Join(m.ritPath, "metrics")
+		if !m.FileWriteReadExister.Exists(path) {
```
brunasilvazup, line 41: Could this `!m.FileWriteReadExister.Exists(path)` be changed to `m.Checker.Check()`?

victor-schumacher: In line 41 no, because the checks function returns the same result if the file does not exist and if the content is "no".
But in line 60 the adjustments are going very well.

pkg/metrics/checker.go
victor-schumacher: Sure! I was worried about having a function on metric.go because it's like a data file, so now I'm thinking about use something like this: ```var FilePath = filepath.Join(api.RitchieHomeDir(), "metrics")```

brunasilvazup: I think it is cool, the important thing is to centralize this ready information.

PR #423, opened 2020-08-13
Asking for credential during formula runtime

henriquemoraes8:
## Description
When a user executes a formula that requires credentials, it halts demanding her to set the credentials through the `rit set credential` command. This behavior may affect the user experience especially in case of custom credentials. 
The idea is to detect the credential keys that have not been set and ask for them during runtime. 

## Discussion
A user may never require to understand how credentials work in Ritchie if he just fills correctly when required. This feature works as an abstraction for those who just want to use the cli without worrying too much about its mechanics.
This will work even for custom credentials or multi-field providers. It records the prompted value as if the user run `rit set credential`
It **does not** create the provider in case it does not exist. The user should manually set a custom provider.
This will record even if the `config.json` file has a typo on either the provider or the key

## How to verify it
Run a formula that requires credentials without having them set

## Description for the changelog
Credentials are asked during formula runtime in case they were not set

The example below shows the behavior running `rit aws create bucket` without previously having any credential set
Screen Shot 2020-08-12 at 22 03 15

### DIFF
--- a/pkg/env/envcredential/env_credential.go
+++ b/pkg/env/envcredential/env_credential.go
@@ -55,11 +55,7 @@ func (c CredentialResolver) Resolve(name string) (string, error) {
 	return credValue, nil
 }
 
-func (c CredentialResolver) PromptCredential(
-	provider string,
-	key string,
-	credentialDetail credential.Detail,
-) (string, error) {
+func (c CredentialResolver) PromptCredential(provider, key string, credentialDetail credential.Detail) (string, error) {
 	inputVal, err := c.Password(
 		fmt.Sprintf("Provider key not found, please provide a value for %s %s: ", provider, key),
 	)
--- a/pkg/env/envcredential/env_credential.go
+++ b/pkg/env/envcredential/env_credential.go
@@ -27,14 +27,15 @@ import (
 type CredentialResolver struct {
 	credential.Finder
 	credential.Setter
+	prompt.InputPassword
 }
 
 // NewResolver creates a credential resolver instance of Resolver interface
-func NewResolver(cf credential.Finder, cs credential.Setter) CredentialResolver {
-	return CredentialResolver{cf, cs}
+func NewResolver(cf credential.Finder, cs credential.Setter, passwordInput prompt.InputPassword) CredentialResolver {
+	return CredentialResolver{cf, cs, passwordInput}
 }
 
-func (c CredentialResolver) Resolve(name string, passwordInput prompt.InputPassword) (string, error) {
+func (c CredentialResolver) Resolve(name string) (string, error) {
 	s := strings.Split(strings.ToLower(name), "_")
 	service := s[1]
 	key := s[2]
@@ -42,12 +43,12 @@ func (c CredentialResolver) Resolve(name string, passwordInput prompt.InputPassw
 	if err != nil {
 		// Provider was never set
 		cred.Service = service
-		return c.PromptCredential(service, key, cred, passwordInput)
+		return c.PromptCredential(service, key, cred)
 	}
 	credValue, exists := cred.Credential[key]
 	if !exists {
 		// Provider exists but the expected key doesn't
-		return c.PromptCredential(service, key, cred, passwordInput)
+		return c.PromptCredential(service, key, cred)
 	}
 
 	// Provider and key exist
@@ -58,9 +59,8 @@ func (c CredentialResolver) PromptCredential(
 	provider string,
 	key string,
 	credentialDetail credential.Detail,
-	passwordInput prompt.InputPassword,
 ) (string, error) {
-	inputVal, err := passwordInput.Password(
+	inputVal, err := c.Password(
 		fmt.Sprintf("Provider key not found, please provide a value for %s %s: ", provider, key),
 	)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/env/envcredential/env_credential_test.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	fileManager := stream.NewFileManager()
+	contextFinder := rcontext.NewFinder(tempDirectory, fileManager)
+	credentialSetter := credential.NewSetter(tempDirectory, contextFinder)
+	credentialFinder := credential.NewFinder(tempDirectory, contextFinder, fileManager)
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			credentialResolver := NewResolver(credentialFinder, credentialSetter, passwordMock{tt.output})
```
brunasilvazup, line 61: Hi @henriquemoraes8,
Here you can use the existing password mock: `inputPasswordMock`.

pkg/env/envcredential/env_credential_test.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+}
+
+type passwordMock struct {
+	value string
+}
+
+func (pass passwordMock) Password(string) (string, error) {
+	return pass.value, nil
+}
```
brunasilvazup, line 80: And using the mock you can remove these lines.

pkg/env/envcredential/env_credential.go
```diff
@@ -27,29 +26,52 @@ import (
...
+	}
+
+	// Provider and key exist
+	return credValue, nil
+}
+
+func (c CredentialResolver) PromptCredential(
+	provider string,
+	key string,
+	credentialDetail credential.Detail,
```
brunasilvazup, line 61: All arguments can be on the same line and also the `provider` and the `key` can divide the `string`, something like `PromptCredential(provider, key string, credentialDetail credential.Detail)`.

Following the pattern (https://tour.golang.org/basics/5)

pkg/env/envcredential/env_credential_test.go
henriquemoraeszup: I have tried, but apparently I cannot use it since it is from a different package in a test file. It seems I can only use structs on regular files from other packages, or am I doing something wrong? 

Besides the current implementation does not suffice, I would have to change it since I need to manipulate the value returned

pkg/env/envcredential/env_credential_test.go
henriquemoraeszup: Same argument as above

pkg/env/envcredential/env_credential.go
henriquemoraeszup: Didn't know that, thanks!

pkg/env/envcredential/env_credential_test.go
henriquemoraeszup: The possible solutions as listed in the article would be to either move these mocks to a `testutil` file or move them to a regular one
https://stackoverflow.com/questions/38712120/how-to-reuse-test-code-in-imported-package

brunasilvazup: Yes, I tested it and it would really be necessary to extract it into a separate package.
And that is not the focus of your PR.

I think you can leave that for now and put it on the backlog to refactor.

pkg/env/envcredential/env_credential_test.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	"os"
+	"testing"
+
+	"github.com/ZupIT/ritchie-cli/pkg/credential"
+	"github.com/ZupIT/ritchie-cli/pkg/rcontext"
+	"github.com/ZupIT/ritchie-cli/pkg/stream"
+)
+
+func TestCredentialResolver(t *testing.T) {
```
brunasilvazup, line 28: A question about the tests, they are not entering the corresponding ifs.

I downloaded his pr, put some stopping points in the ifs and saw that when running the test in debug mode it does not stop.

I believe that it is through the `credentialFinder` that you are going through, and also because you are creating the file outside the tests, that is, it uses the same file (and preserves the information) in the three tests.

brunasilvazup: In the case of mocking the Finder you can use examples from `mocks_test.go` and create in the file itself as it will be the only one currently used.

henriquemoraeszup: Hi @brunasilvazup the three cases are all covered. Preserving the info between tests was intentional
* On the first the file must be created
* On the second, the file exists but does not have a key
* On the third the key has already been written

Not mocking the finder is important here to effectively write and read the credential file from a temp directory

Please check the coverage file to validate that the cases are accounted for
https://codecov.io/gh/ZupIT/ritchie-cli/pull/423/diff?src=pr&el=tree#diff-cGtnL2Vudi9lbnZjcmVkZW50aWFsL2Vudl9jcmVkZW50aWFsLmdv

Any questions let me know

brunasilvazup: As we talked the test was running correctly only the first time.

henriquemoraeszup: Corrected as discussed in the call 🚀

PR #425, opened 2020-08-13
Add functional test for list

DittrichLucas:
**- What I did**
- Add functional test for core `list` command
- Fix lint

**- How to verify it**
- Run: `make functional-test`

- Add functional test for core list command and fix lint

### DIFF
--- a/functional/core/core_integration_test.go
+++ b/functional/core/core_integration_test.go
@@ -52,21 +52,22 @@ var _ = Describe("RitCore", func() {
 		Entry("Add", scenariosCore[3]),
 		// Entry("Add new repo", scenariosCore[4]),
 
-		Entry("List repo", scenariosCore[5]),
-		// Entry("Delete repo", scenariosCore[6]),
+		Entry("List", scenariosCore[5]),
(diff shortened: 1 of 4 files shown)


functional/core/core_integration_test.go
```diff
@@ -52,21 +52,22 @@ var _ = Describe("RitCore", func() {
 		Entry("Add", scenariosCore[3]),
 		// Entry("Add new repo", scenariosCore[4]),
 
-		Entry("List repo", scenariosCore[5]),
-		// Entry("Delete repo", scenariosCore[6]),
+		Entry("List", scenariosCore[5]),
```
brunasilvazup, line 55: I've been thinking, couldn't we use `` instead of writing the same thing?

I know it is less explicit, but it is less easy to make mistakes too.

lucasdittrichzup: I agree with you, it's redundant! I'm going to change the file I'm moving and we can see how to change the others

lucasdittrichzup:
This PR will be opened to the `master`.

PR #429, opened 2020-08-13
Collector function for metrics

victor-schumacher:
**- What I did**
Added a Data collector function which gonna be used on metrics.

**- How to verify it**
Use the function or check the unit tests.

Data collector function for metrics


pkg/metrics/metric.go
```diff
@@ -27,17 +27,29 @@ var FilePath = filepath.Join(api.RitchieHomeDir(), "metrics")
...
+type Data struct {
+	UserId       UserId `json:"userId"`
+	OS           OS     `json:"operationalSystem:"`
+	Command      Command `json:"command"`
+	CommandError CommandError `json:"commandError"`
+}
+
 type Sender interface {
-	Send(dataset Dataset)
+	Send(dataset Metric)
```
brunasilvazup, line 52: Here you could also change the variable name for metrics?

pkg/metrics/metric.go
```diff
@@ -27,17 +27,29 @@ var FilePath = filepath.Join(api.RitchieHomeDir(), "metrics")
 
 type Id string
 
-type UserId string
-
-type Dataset struct {
+type Metric struct {
```
kaduartur, line 30: maybe this name doesn't make sense, because the package is already called metric.
could you add the `RitVersion` too?

pkg/metrics/metric.go
```diff
@@ -27,17 +27,29 @@ var FilePath = filepath.Join(api.RitchieHomeDir(), "metrics")
...
 
+type UserId string
+
+type OS string
+
+type Command string
+
+type CommandError string
+
+type Data struct {
```
kaduartur, line 44: UserId and OS should are into Metric struct to respect the API contract

pkg/metrics/data_collector.go
```diff
@@ -0,0 +1,62 @@
...
+
+	data := Data{
+		UserId:  userId,
+		OS:      OS(runtime.GOOS),
+		Command: Command(joinArgs(" ")),
+		CommandError: CommandError(strings.Join(commandError, " ")),
+	}
+
+	metric := Metric{
+		Id:        Id(uuid.New().String()),
```
kaduartur, line 51: The metricId should be the command to facilitate the queries and joins in the database

pkg/metrics/data_collector.go
```diff
@@ -0,0 +1,62 @@
...
+		Timestamp: time.Now(),
+		Data:      data,
+	}
+
+	return metric, nil
+}
+
+func joinArgs(sep string) string {
```
kaduartur, line 59: Could you create a more specific name for this function?

pkg/metrics/metric.go
victor-schumacher: Changed to APIData

PR #430, opened 2020-08-13
Add functional test for list

DittrichLucas:
**- What I did**
- Add functional test for core `list` command
- Fix lint

**- How to verify it**
- Run: `make functional-test`

- Add functional test for core list command and fix lint

### DIFF
--- a/functional/core/core_integration_test.go
+++ b/functional/core/core_integration_test.go
@@ -45,28 +45,28 @@ var _ = Describe("RitCore", func() {
 			Expect(out).To(ContainSubstring(scenario.Result))
 		},
 
-		Entry("Show context", scenariosCore[0]),
-		// Entry("Set context", scenariosCore[1]),
-		// Entry("Delete context", scenariosCore[2]),
+		Entry(scenariosCore[0].Entry, scenariosCore[0]),
+		// Entry(scenariosCore[1].Entry, scenariosCore[1]),
+		// Entry(scenariosCore[2].Entry, scenariosCore[2]),
 
-		Entry("Add", scenariosCore[3]),
-		// Entry("Add new repo", scenariosCore[4]),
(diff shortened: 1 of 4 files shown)


functional/core/core_integration_test.go
```diff
@@ -45,28 +45,28 @@ var _ = Describe("RitCore", func() {
 			Expect(out).To(ContainSubstring(scenario.Result))
 		},
 
-		Entry("Show context", scenariosCore[0]),
-		// Entry("Set context", scenariosCore[1]),
-		// Entry("Delete context", scenariosCore[2]),
+		Entry(scenariosCore[0].Entry, scenariosCore[0]),
+		// Entry(scenariosCore[1].Entry, scenariosCore[1]),
+		// Entry(scenariosCore[2].Entry, scenariosCore[2]),
```
victor-schumacher, line 50: these comments are used to something?

functional/runner.go
```diff
@@ -50,7 +50,7 @@ func (scenario *Scenario) RunSteps() (string, error) {
 	fmt.Println("Running: " + scenario.Entry)
 
 	os := runtime.GOOS
-	if  os == "windows" && len(scenario.Steps) >= 2 {
+	if os == "windows" && len(scenario.Steps) >= 2 {
```
victor-schumacher, line 53: maybe is a good idea to create a const called windows, so it gets easier to change

lucasdittrichzup: yes, it's a good idea, but i'll solve this in the next pr

functional/core/core_integration_test.go
lucasdittrichzup: yes, these tests are only commented because they are failing and, therefore, break the build. As it is something old, I preferred to create the tests for the commands that do not yet have and leave them to analyze when adjusting the tests for windows

PR #450, opened 2020-08-19
Add collect metric to the main file

kaduartur:
**- What I did**
- Added collect metrics to the main file

**- How to verify it**
Run any commands on the rit

Add collect metrics and send metrics to server

### DIFF
--- a/pkg/cmd/metrics_test.go
+++ b/pkg/cmd/metrics_test.go
@@ -101,23 +101,6 @@ func Test_metricsCmd_runPrompt(t *testing.T) {
 			wantErr: false,
 		},
 		{
-			name: "fail on check when metrics file exist",
-			in: in{
-				file: sMocks.FileWriteReadExisterCustomMock{
-					ExistsMock: func(path string) bool {
-						return true
-					},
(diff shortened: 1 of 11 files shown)


pkg/cmd/metrics_test.go
```diff
@@ -111,11 +111,11 @@ func Test_metricsCmd_runPrompt(t *testing.T) {
 						return nil
 					},
 				},
-				checkerMock: CheckerMock{func() (bool, error) {
-					return false, errors.New("error reading file")
+				checkerMock: CheckerMock{func() bool {
+					return false
```
brunasilvazup, line 115: In that case, as there will be no more errors returned from Check, I think you can remove this test no?

pkg/metric/checker_test.go
```diff
@@ -79,7 +75,6 @@ func Test_Check(t *testing.T) {
 		},
 		{
 			name:           "fail case error on reading file",
```
brunasilvazup, line 77: What do you think of switching to `success case expecting false when error reading file`

Issue #390, opened 2020-08-10
Conditional options on config.json


rasouza:
**What would you like to be added**:
I'd like to show an option based on the answer of the previous option
**Why is this needed**:
So I can deliver a better experience to the user by removing unnecessary questions depending on his/her answer

victor-schumacher:
@rasouza that sound's nice! Thanks for the idea.
Today cli reads inputs from config.json on this file:
https://github.com/ZupIT/ritchie-cli/blob/4d0af9c251e9b28139694df0458caa53d65da8f3/pkg/formula/runner/inputs.go#L115

The implementation of conditional options will probably go there.
Pull requests about this feature are welcome.

henriquemoraeszup:
Hey @rasouza ! thanks for your suggestion. Your request is currently under review on #454.
Stay calm, and automate everything 👊

PR #454, opened 2020-08-21
Adding conditional steps to config prompt

henriquemoraeszup:
## Description
As requested on issue #390, this PR allows the developer to specify optional inputs according to previous user inputs. The user needs to create a key called `condition` on an optional step and provide a variable reference from any previous variable defined, an operator for comparison (==, !=, >, >=, <, <=), and a value to compare. If the condition exists and fails, the current step is not shown.

This is a simple implementation some of the points taken in account are explained below:

* The comparison operators are implemented in a `case` like manner. It is not very elegant. The better approach would be to evaluate the expression, but it was not chosen because:
    1. Go does not come with a native `eval` function and it would require to add a dependency to the project for a simple task
    2. Since many times the user input is free we could risk him adding harmful code as a value and executing it on an `eval` command or letting the author add malicious code on the `operator` variable
* This approach only attends simple expressions between a single variable and a value. Complex multi-variable expressions are not supported for the sake of avoiding overengineering. This should cover most of the cases, I think it is better to wait for the community to demand more complex interactions before we actually implement them.
* There are discussions to create tree-like decisions to prompt the user. This implementation does not rule that out, we can still implement a decision tree with conditional nodes, so forward-compatibility is accounted for.
* In case the user uses a wrong operator. An error is thrown
* In case the user defines a non existing variable name. An error is thrown

Suggestions are welcome!

## How to verify it
Create a condition key on `config.json` of a formula like on this example
```
{
      "cache": {
        "active": true,
        "newLabel": "Type new value. ",
        "qty": 6
      },
      "label": "Type : ",
      "name": "sample_text",
      "condition": {
          "variable": "sample_list",
          "operator": "==",
          "value": "in_list1"
      },
      "type": "text"
    },
```
Run the formula and **fail the condition** to have this prompt not shown

## Description for the changelog
Adding conditional steps to config prompt

Aug-20-2020 21-27-16

### DIFF
--- a/pkg/formula/runner/inputs.go
+++ b/pkg/formula/runner/inputs.go
@@ -121,6 +121,10 @@ func (in InputManager) fromPrompt(cmd *exec.Cmd, setup formula.Setup) error {
 		if err != nil {
 			return err
 		}
+		if !in.verifyConditional(cmd, input) {
+			continue
+		}
+
 		switch iType := input.Type; iType {
 		case "text":
 			if items != nil {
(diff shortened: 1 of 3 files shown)


pkg/formula/runner/inputs.go
```diff
@@ -246,3 +250,43 @@ func (in InputManager) resolveIfReserved(input formula.Input) (string, error) {
...
+	// optimizations are welcome, being mindful of the points above
+	switch input.Condition.Operator {
+	case "==": return value == input.Condition.Value
+	case "!=": return value != input.Condition.Value
+	case ">":  return value > input.Condition.Value
+	case ">=": return value >= input.Condition.Value
+	case "<":  return value < input.Condition.Value
+	case "<=": return value <= input.Condition.Value
+	default:
+		message := fmt.Sprintf(
```
kaduartur, line 285: you should return an error when the operator is invalid, because the fallback to equality may not be clear to the user.
Taking into account programming languages, if you pass an invalid operator your code will not compile.

pkg/formula/runner/inputs.go
```diff
@@ -246,3 +250,43 @@ func (in InputManager) resolveIfReserved(input formula.Input) (string, error) {
...
+	variable := input.Condition.Variable
+	for _, envVal := range cmd.Env {
+		components := strings.Split(envVal, "=")
+		if strings.ToLower(components[0]) == variable {
+			value = strings.ToLower(components[1])
+			break
+		}
+	}
+	if value == "" {
+		message := fmt.Sprintf("config.json: Conditional variable %s not found. Showing prompt item anyway.", variable)
```
kaduartur, line 269: You should return an error if the condition variable does not exist.
Consider the same reasons as the comment below.

henriquemoraeszup: You got it chief

rasouza:
Amazing job indeed 👏 

One further improvement suggestion: we could implement a parser to get rid of the `condition` object and let it be written as simples as
```
...
"condition": "sample_list == in_list1"
...
```

This implementation of yours is great already, though. Thanks!

henriquemoraeszup:
Actually the parser is a good idea, it would allow us to cover the point above on multi-variables as well. We can leave as an improvement =]

> Amazing job indeed 👏
> 
> One further improvement suggestion: we could implement a parser to get rid of the `condition` object and let it be written as simples as
> 
> ```
> ...
> "condition": "sample_list == in_list1"
> ...
> ```
> 
> This implementation of yours is great already, though. Thanks!

PR #465, opened 2020-08-24
Add functional test for tutorial

DittrichLucas:
**- What I did**
- Add functional test for core `tutorial` command

**- How to verify it**
- Run: `make functional-test`

- Add functional test for core tutorial command

### DIFF
--- a/functional/core/core_feature.json
+++ b/functional/core/core_feature.json
@@ -317,5 +317,37 @@
             }
         ],
         "result":"Use \"rit <command> --help\" for more information about a given command."
+    },
+    {
+        "entry":"Tutorial enable",
+        "steps":[
+            {
+                "key":"",
+                "value":"tutorial",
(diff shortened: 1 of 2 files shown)


functional/core/core_feature.json
```diff
@@ -317,5 +317,37 @@
...
+                "action":"main"
+            },
+            {
+                "key":"Status tutorial?",
+                "value":"enabled",
+                "action":"select"
+            }
+        ],
+        "result":"Set tutorial successful!"
+    },
```
brunasilvazup, line 336: Hi Lucas
There, in the tutorial tests, wouldn't an option be enough?

Because the intention is to test the output and it does not change with the options.
What do you think?

lucasdittrichzup: I agree with you and I already submit the change by removing one of the tests: D

PR #478, opened 2020-08-27
Added step to question about formula found

brunats:
**- What I did**
Added a new step in choosing the formula for the build.

When rit finds a directory with a formula, it looks for other formulas in the tree (within the same directory).
If he finds it, he asks the user if he wants to select this formula found or continue searching for another one.

> Other formulas found and the first selected:
> rit-build-formula-1

>Found other formulas and selected to continue searching:
> rit-build-formula-2

**- How to verify it**
> Clone this pr with the command and build Ritchie
```bash
git fetch upstream pull/478/head:fix/order-of-choice-of-formula-for-the-build
git checkout fix/order-of-choice-of-formula-for-the-build
make build && sudo cp dist/linux/rit /usr/local/bin/
```
> And run rit add repo
```bash
rit create formula # using 'rit group verb'
rit create formula # using 'rit group verb formula'
rit build formula # select rit group verb formula
rit build formula # select rit group verb
```

Added step to question about formula founded.

### DIFF
--- a/pkg/cmd/build_formula.go
+++ b/pkg/cmd/build_formula.go
@@ -33,9 +33,10 @@ import (
 )
 
 const (
-	newWorkspace = "Type new formula workspace?"
-	docsDir      = "docs"
-	srcDir       = "src"
+	newWorkspace               = "Type new formula workspace?"
+	docsDir                    = "docs"
+	srcDir                     = "src"
+	questionSelectFormulaGroup = "Select a formula or group: "
 )
 
 type buildFormulaCmd struct {
@@ -180,7 +181,7 @@ func (b buildFormulaCmd) readFormulas(dir string, currentFormula string) (string
 		dirs = sliceutil.Remove(dirs, srcDir)
 	}
 
-	selected, err := b.List("Select a formula or group: ", dirs)
+	selected, err := b.List(questionSelectFormulaGroup, dirs)
 	if err != nil {
 		return "", err
 	}
--- a/pkg/cmd/build_formula.go
+++ b/pkg/cmd/build_formula.go
@@ -33,10 +33,11 @@ import (
 )
 
 const (
-	newWorkspace               = "Type new formula workspace?"
-	docsDir                    = "docs"
-	srcDir                     = "src"
-	questionSelectFormulaGroup = "Select a formula or group: "
+	newWorkspace                = "Type new formula workspace?"
+	docsDir                     = "docs"
+	srcDir                      = "src"
+	questionSelectFormulaGroup  = "Select a formula or group: "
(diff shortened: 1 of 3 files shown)


pkg/cmd/build_formula_test.go
```diff
@@ -200,10 +186,29 @@ func TestBuildFormulaCmd(t *testing.T) {
...
+	}
+	if fieldsTest.localBuilder != nil {
+		fields.localBuilder = fieldsTest.localBuilder
+	}
+	if fieldsTest.workspaceManager != nil {
+		fields.workspaceManager = fieldsTest.workspaceManager
+	}
+
+	return fields
+}
```
brunasilvazup, line 282: This function is used to always choose the default fields and if the test defines any field, use the test field.

The https://github.com/ZupIT/ritchie-cli/pull/478/commits/d5b81bc392cdb8705570c56026cd202cc7632311 commit is the one with just that change.

pkg/cmd/mocks_test.go
```diff
@@ -397,6 +397,58 @@ func (t TutorialFindSetterCustomMock) Set(tutorial string) (rtutorial.TutorialHo
 	return t.set(tutorial)
 }
 
+type DirManagerCustomMock struct {
+	exists func(dir string) bool
+	list   func(dir string, hiddenDir bool) ([]string, error)
+	isDir  func(dir string) bool
+}
```
victor-schumacher, line 404: this will be very useful to test thinks like formula creator, watcher or template manager

pkg/cmd/build_formula.go
```diff
@@ -151,24 +154,41 @@ func (b buildFormulaCmd) build(workspacePath, formulaPath string) {
...
+			return "", err
+		}
+		if response == currentFormula {
+			return dir, nil
+		}
+		dirs = sliceutil.Remove(dirs, srcDir)
 	}
 
-	selected, err := b.List("Select a formula or group: ", dirs)
+	selected, err := b.List(questionSelectFormulaGroup, dirs)
```
lucasdittrichzup, line 185: in your GIF it seems that we have a question for **group** and another for **verb**, but in the description of the question there is **only group**, don't you think this will cause confusion for users?

brunasilvazup: In the gif the questions are always the same "Select a formula or group:", but the group and verb used are the composition of the formula created.

The formulas that were formulas for example were: `rit group verb` and `rit group verb formula`, choices that may have caused your confusion.

brunasilvazup:
\merge qa

brunasilvazup:
Qa tests:
 - two formulas with the same starting structure;
 - build tests of both;
- created normal formula and tested the build with the creation.

PR #487, opened 2020-08-31
add metrics question on rit upgrade command

victor-schumacher:
**- What I did**
Added a step on rit upgrade, which checks for metrics file on .rit, if doesn't exist, rit asks about metrics and create the file.

Now api.RitchieHomeDir() returns the same dir when command are called with sudo

**- How to verify it**
delete your metrics file, run rit upgrade

Added metrics question on rit upgrade.


lucasdittrichzup (approved):

brunasilvazup (approved):

pkg/api/api.go
```diff
@@ -97,12 +98,23 @@ func (t TermInputType) ToLower() string {
 }
 
 // UserHomeDir returns the home dir of the user
+// return usr.HomeDir
```
lucasdittrichzup, line 101: in golang, it is good practice to subscribe to a function that can be exported starting with the function name itself. In that case, it would be UserHomeDir ...

lucasdittrichzup (approved):

pkg/api/api.go
victor-schumacher: it happens on line 100 no?

victor-schumacher: i'm refactoring this comment to be more clear

kaduartur (approved):
LGTM!

PR #496, opened 2020-09-03
Remove command lock without initialization

brunats:
**- What I did**

Removed the block on using commands when not adding a commons repository.

Now only the `rit create formula` command will remain blocked, other core commands or commands added via the repository will function normally.

>aa

**- How to verify it**

> Clone this pr with the command and build Ritchie
```bash
git fetch upstream pull/496/head:improvement/change-command-locking-without-rit-init
git checkout improvement/change-command-locking-without-rit-init
make build && sudo cp dist/linux/rit /usr/local/bin/
```
> And run rit add repo
```bash
rit list repo # It is not blocked
rit create formula # Is blocked
```

Remove blocking of most commands when no commons repository is added


pkg/cmd/root.go
```diff
@@ -62,6 +62,10 @@ var (
 	upgradeList = []string{
 		cmdUse,
 	}
+
+	listCommandsBlockedByCommons = []string{
```
kaduartur, line 66: Could you rename the variable to `blockedCmdsByCommons`?
What do you think about it?

kaduartur (changes requested):
LGTM! 🚀

pkg/cmd/root.go
brunasilvazup: I think it's a good idea.

Makefile
```diff
@@ -118,7 +118,10 @@ unit-test:
 
 functional-test:
 	mkdir -p $(BIN)
-	$(GO_TEST) -v -count=1 -p 1 `go list ./functional/... | grep -v vendor/`
+	$(GO_TEST) -v -count=1 -p 1 `go list ./functional/init/... | grep -v vendor/`
```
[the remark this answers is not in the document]
brunasilvazup: I just don't know if it would be clear the need for the core tests to be last, but I think it is a valid solution today.

kaduartur (approved):

victor-schumacher (approved):

lucasdittrichzup (approved):

PR #499, opened 2020-09-04
add check for http status

victor-schumacher:
**- What I did**

Add a http status check when request for stable version

**- How to verify it**

changing the stable version url for some url which don't return 200 and run ``` rit upgrade ```

Add a http status check when request for stable version

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -72,8 +72,8 @@ func requestStableVersion(stableVersionUrl string, httpClient *http.Client) (str
 	}
 
 	response, err := httpClient.Do(request)
-
-	if err != nil || response.StatusCode != http.StatusOK {
+	//  || response.StatusCode != http.StatusOK
+	if err != nil {
 		return "", err
 	}
 
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -2,6 +2,7 @@ package version
 
 import (
 	"encoding/json"
+	"errors"
 	"io/ioutil"
 	"net/http"
 	"strings"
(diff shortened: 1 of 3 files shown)


lucasdittrichzup (approved):

brunasilvazup (approved):
You have some functional tests of teams breaking in the build, but otherwise the functionality is ok.

victor-schumacher:
i commented my changes to test the only pipeline, and it breaks

kaduartur:
> I commented my changes to test the only pipeline, and it breaks

The ritchie-server (https://ritchie-server.itiaws.dev) is down, try again later!

victor-schumacher:
> > I commented my changes to test the only pipeline, and it breaks
> 
> The ritchie-server (https://ritchie-server.itiaws.dev) is down, try again later!
Hi @kaduartur, i tried now but still failing, is there any specific time?

victor-schumacher:
> > > I commented my changes to test the only pipeline, and it breaks
> > 
> > 
> > The ritchie-server (https://ritchie-server.itiaws.dev) is down, try again later!
> > Hi @kaduartur, i tried now but still failing, is there any specific time?

Since the QA environment used on team's functional test was removed, we are removing this step from pipeline too.

pkg/version/version.go
```diff
@@ -72,10 +72,10 @@ func requestStableVersion(stableVersionUrl string, httpClient *http.Client) (str
 	}
 
 	response, err := httpClient.Do(request)
-
-	if err != nil {
+	if err != nil || response.StatusCode != http.StatusOK {
```
kaduartur, line 75: I think that when StatusCode isn't 200 the err can be nil and return this not make sense. 

Could you create a new error type and create a new test for this case?

kaduartur (changes requested):

pkg/version/version.go
victor-schumacher: Thanks for the advice, that's so true.
Done.

kaduartur (approved):

PR #512, opened 2020-09-10
Add validation with regex in the config.json inputs

JoaoDanielRufino:
Signed-off-by: JoaoDanielRufino <joaodaniel0405@gmail.com>

**- What I did**
Added regex type to config.json to check if user input is in accordance with he specified regex in config.json, if not, it will print the mismatchText message and wait for a valid input

**- How to verify it**
Inside config.json include: 
```
{
  "dockerImageBuilder": "cimg/base:stable-20.04",
  "inputs": [
    {
      "cache": {
        "active": true,
        "newLabel": "Type new value: ",
        "qty": 6
      },
      "label": "Cluster Name: ",
      "name": "cluster_name",
      "type": "text",
      "pattern": {
            "regex": "your_regex",
            "mismatchText": "your mismatch text"
      }
    }
  ]
}
```
Regex type to config.json


pkg/formula/runner/inputs.go
```diff
@@ -134,12 +139,17 @@ func (in InputManager) fromPrompt(cmd *exec.Cmd, setup formula.Setup) error {
 			if items != nil {
 				inputVal, err = in.loadInputValList(items, input)
 			} else {
-				validate := input.Default == ""
-				inputVal, err = in.Text(input.Label, validate)
-				if inputVal == "" {
-					inputVal = input.Default
+				if len(input.Pattern.Regex) > 0 {
+					inputVal, err = in.textRegexValidator(input)
```
victor-schumacher, line 143: What you think about a validate function? The same logic is written on 221.

pkg/formula/runner/inputs.go
```diff
@@ -208,10 +218,14 @@ func (in InputManager) loadInputValList(items []string, input formula.Input) (st
...
 	if inputVal == newLabel {
-		validate := len(input.Default) == 0
-		inputVal, err = in.Text(input.Label, validate)
-		if len(inputVal) == 0 {
-			inputVal = input.Default
+		if len(input.Pattern.Regex) > 0 {
+			inputVal, err = in.textRegexValidator(input)
+		} else {
+			validate := len(input.Default) == 0
+			inputVal, err = in.InputText.Text(input.Label, validate)
```
victor-schumacher, line 225: It seems like the err varible is overwritten on 225, is it used for anything?

pkg/formula/runner/inputs.go
```diff
@@ -277,16 +291,32 @@ func (in InputManager) verifyConditional(cmd *exec.Cmd, input formula.Input) (bo
...
+	case "!=":
+		return value != input.Condition.Value, nil
+	case ">":
+		return value > input.Condition.Value, nil
+	case ">=":
+		return value >= input.Condition.Value, nil
+	case "<":
+		return value < input.Condition.Value, nil
+	case "<=":
+		return value <= input.Condition.Value, nil
```
victor-schumacher, line 305: nice, for me it got much more readable

pkg/formula/runner/inputs.go
joaorufinozup: I have just followed the logic that was written before. It was being overwritten to just return at the end of the function

victor-schumacher: Sure, what you think about check the error or ignore it?

joaorufinozup: I am going to check it

pkg/formula/runner/inputs.go
```diff
@@ -134,13 +139,20 @@ func (in InputManager) fromPrompt(cmd *exec.Cmd, setup formula.Setup) error {
...
-				inputVal, err = in.Text(input.Label, validate, input.Tutorial)
-
-				if inputVal == "" {
-					inputVal = input.Default
+				if in.hasRegex(input) {
+					inputVal, err = in.textRegexValidator(input)
+					if err != nil {
+						return err
+					}
+				} else {
```
kaduartur, line 147: could you remove `else` statement and add `break`?

pkg/formula/runner/inputs.go
```diff
@@ -209,10 +221,14 @@ func (in InputManager) loadInputValList(items []string, input formula.Input) (st
 	}
 	inputVal, err := in.List(input.Label, items, input.Tutorial)
 	if inputVal == newLabel {
-		validate := len(input.Default) == 0
-		inputVal, err = in.Text(input.Label, validate, input.Tutorial)
-		if len(inputVal) == 0 {
-			inputVal = input.Default
+		if in.hasRegex(input) {
+			inputVal, err = in.textRegexValidator(input)
```
kaduartur, line 225: could you return `in.textRegexValidator(input)`and remove `else` statement?

pkg/formula/runner/inputs_test.go
```diff
@@ -432,6 +446,69 @@ func TestInputManager_ConditionalInputs(t *testing.T) {
 	}
 }
 
+func TestInputManager_RegexType(t *testing.T) {
```
kaduartur, line 449: Could you create a table test to following the Golang tests standard and add more field for **inputJson?**

kaduartur (changes requested):
Waiting for your changes. Good job!

pkg/formula/runner/inputs.go
joaorufinozup: I did not understand, because if the break the for loop will end, is this the correct behavior that you want?

pkg/formula/runner/inputs.go
```diff
@@ -134,13 +139,20 @@ func (in InputManager) fromPrompt(cmd *exec.Cmd, setup formula.Setup) error {
...
+					inputVal, err = in.textRegexValidator(input)
+					if err != nil {
+						return err
+					}
+				} else {
+					validate := isRequired(input)
+					inputVal, err = in.InputText.Text(input.Label, validate, input.Tutorial)
+					if inputVal == "" {
+						inputVal = input.Default
+					}
```
brunasilvazup, line 152: Hello @JoaoDanielRufino ,

I was testing the regex with the required attribute and it didn't work well.
If I have an input that is optional (required == false) and a regex, I can't pass an empty string (which would be not to send the same one).

And thinking about a solution I saw that the regex (when existing) would be another validator of the input string.

I thought about changing the name of the function `textRegexValidator` to` textValidator` and putting in it all the needs of this string entry:

 - a part validating the required
 - Other part of the regex itself (with or without the need for it - 
Using else if there `if in.hash Regex (input) {`


Similar to https://github.com/ZupIT/ritchie-cli/pull/544/files in the`pkg/cmd/create_formula.go` file and function `formulaCommandValidator`.

If you want to talk about this solution just let me know.

brunasilvazup (approved):
LGTM

pkg/formula/runner/docker/runner_test.go
```diff
@@ -220,6 +220,12 @@ func (e envResolverMock) Resolve(string) (string, error) {
 	return e.in, e.err
 }
 
+type inputTextValidatorMock struct{}
+
+func (inputTextValidatorMock) Text(name string, validate func(interface{}) error, helper ...string) (string, error) {
+	return "mocked text", nil
```
henriquemoraeszup, line 226: can we have a dynamic parameter here?

henriquemoraeszup (approved):

kaduartur (approved):

PR #517, opened 2020-09-11
Added commons repository acceptance info to metrics

victor-schumacher:
**- What I did**

Added a new bool field to metrics struct
Little refactor at init strings

**- How to verify it**

Use `rit init` , accept, or not the metrics question, then check for metrics on metabse, or log the struct.

Added a new bool field to metrics struct


lucasdittrichzup (approved):

brunasilvazup (approved):
:+1:

PR #520, opened 2020-09-14
Create formulas with internal templates

kaduartur:
**- What I did**

- I created templates.go file with the [go-bindata](https://github.com/go-bindata/go-bindata)
- I modified the `formula/creator/creator.go` and `formula/creator/template/manager.go` to use the templates.go file to create new formulas.
- Block commands when the rit has not been initialized, just the allow list can be run.

**- How to verify it**
1) Build this branch and execute the `rit init` command.
2) Try to create a new formula with `rit create formula command`

Add formula templates to ritchie-cli


pkg/cmd/root.go
```diff
@@ -185,6 +184,12 @@ func tutorialRit(tutorialStatus string) {
 }
 
 func (ro *rootCmd) ritchieIsInitialized() bool {
-	commonsRepoPath := filepath.Join(ro.ritchieHome, "repos", "commons")
-	return ro.dir.Exists(commonsRepoPath)
+	metricsFile := filepath.Join(ro.ritchieHome, "metrics")
```
victor-schumacher, line 187: is possible to use metrics.FilePath here? or it is dep cycle?

victor-schumacher (dismissed):
I didn't review the files inside /template/~
Core files looks good to me. :100:

pkg/cmd/root.go
kaduartur: It's a good way, thanks!

victor-schumacher (approved):

pkg/cmd/root.go
```diff
@@ -113,7 +112,7 @@ func (ro *rootCmd) PreRunFunc() CommandRunnerFunc {
 			return nil
 		}
 
-		if !ro.ritchieIsInitialized() && isBlockedByCommons(blockedCmdsByCommons, cmd) {
+		if !ro.ritchieIsInitialized() && !isAllowList(cmd) {
```
brunasilvazup, line 115: Will we now return most of the blocked commands?
I thought that this pr would not have that influence.

kaduartur: yes, because we are losing metrics for not forcing the user to init. but as soon as he performs the rit init he can do anything, even without adding the repository commons

PR #551, opened 2020-09-23
Add required field to Input struct

kaduartur:
**- What I did**
I added the new field **required** of type boolean to Input struct to validate if the field can be optional

**- How to verify it**
Add the new field `required` to a config.json.
Example:
```json
[
    {
        "name": "sample_text_2",
        "type": "text",
        "label": "Type : ",
	"required": true
    }
]
```

New field to config.json `required`

### DIFF
--- a/pkg/formula/formula.go
+++ b/pkg/formula/formula.go
@@ -50,6 +50,7 @@ type (
 		Cache     Cache     `json:"cache"`
 		Condition Condition `json:"condition"`
 		Tutorial  string    `json:"tutorial"`
+		Required  *bool     `json:"required"`
 	}
 
 	Cache struct {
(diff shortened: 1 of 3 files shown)


brunasilvazup (approved):

GuillaumeFalourd:
Will it still be necessary to check the value inside the formula @kaduartur ?

What I mean is: if the input is coded as `"required":false` and the user doesn't inform any value, the local variable with the input name will be set empty, or it won't be set ?

kaduartur:
> Will it still is necessary to check the value inside the formula @kaduartur ?
> 
> What I mean is: if the input is coded as `"required": false` and the user doesn't inform any value, the local variable with the input name will be set empty, or it won't be set?

@GuillaumeFalourd the environment variable that represents the field will be sent empty to formula

PR #563, opened 2020-09-28
Add sending metrics for the --watch flag

DittrichLucas:
**- What I did**

- Added sending metrics for the `--watch` flag

**- How to verify it**

- Access the metrics dashboard

- add sending metrics for the --watch flag

### DIFF
--- a/pkg/formula/watcher/watcher.go
+++ b/pkg/formula/watcher/watcher.go
@@ -32,15 +32,26 @@ import (
 )
 
 type WatchManager struct {
-	watcher *watcher.Watcher
-	formula formula.LocalBuilder
-	dir     stream.DirListChecker
+	watcher    *watcher.Watcher
+	formula    formula.LocalBuilder
+	dir        stream.DirListChecker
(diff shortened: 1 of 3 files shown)


pkg/formula/watcher/watcher.go
```diff
@@ -70,6 +81,9 @@ func (w *WatchManager) Watch(workspacePath, formulaPath string) {
 	watchText := fmt.Sprintf("Watching dir %s \n", formulaPath)
 	prompt.Info(watchText)
 
+	startTime := time.Now().Second()
+	w.sendMetric(float64(startTime))
```
brunasilvazup, line 85: In case the formula breaks for some mistake, that would send two metrics right?

lucasdittrichzup: no, it will only be executed on the first build. In this first moment we will collect it like this because it would generate a lot of _"dirt"_ if every change (be it with error or not) we send the metric

brunasilvazup (approved):

victor-schumacher (approved):

PR #578, opened 2020-10-02
Prompt to auto rebuild formula on source modifications

### LINKED ISSUE #559: Suggest a build when the formula is used and we notice that a modification was made in the files

When the formula is used and the files are different from the last use, suggest a build option automatically.

Or build automatically (without asking) when a modification is noticed

Ex:

rit clubhouse create story

- We've noticed your files have modifications from last build, do you want to build the formula? 
yes
no

If the answer is yes

Build automatically

**Why is this needed**:

Improve usability. If you are making modifications, is likely you want to use the latest version.

gabriel-pinheiro:
**- What I did**
Now, when there are changes in a formula, Ritchie will ask the user whether he wants to run it as is or recompile it automatically before running

**- How to verify it**
- Run any formula to have its first build generated
- Modify its source files
- Run it again, you'll be asked to rebuild it
rit2

Prompt to auto rebuild formula on source modifications

Closes #559


gabriel-pinheiro:
You can already use the feature in the branch, it was opened as a Draft because:
- Code will be refactored: I'll extract the building part of the pre_run.go to a new file
- Tests: The changes broke the tests, I'll take a look at it later and create tests for the new feature

In parallel, I'd like to get help from the Ritchie team to make some decisions:
- Currently the modifications are detected by the directory checksum being changed.. I'm saving the checksum of each formula in a separate file under ~/.rit/hashes
Do you think this is a great idea or would it be better to save all of them in a single JSON file like ~/.rit/repos/hashes.json ?
I don't think marshall/unmarshall times would be that big even for users with a lot of formulas

- The message is currently:
"This formula has changed since the last run, would you like to rebuild?"
Is that ok or would you suggest something else?

- I'm afraid hashing time will be too long for formulas with too many files (like NodeJS formulas with huge node_modules folders), what do you think is the best solution to prevent this? Skip the hash when file count is above a threshold? Limit the hash to a certain directory deepness? (would detect most changes but not all of them) Limit the hashing time? Or is that me overworring about it? Afterall, formulas aren't generally that big?

Thank you in advance :D

pkg/formula/runner/local/pre_run_build.go
```diff
@@ -0,0 +1,202 @@
...
+	// Test: should always run updateHash, even when getPreviousHash fails
+	previousHash, getErr := this.getPreviousHash(formulaPath)
+	currentHash, setErr := this.updateHash(formulaPath)
+
+	if getErr != nil {
+		return false, nil
+	}
+	if setErr != nil {
+		return false, setErr
+	}
```
victor-schumacher, line 140: i think ```getErr``` can be just ```err``` , and be checked right after the operation.
same for ```setErr```

pkg/formula/runner/local/pre_run_build.go
```diff
@@ -0,0 +1,202 @@
...
+	return previousHash != currentHash, nil
+}
+
+func (this PreRunBuilder) getPreviousHash(formulaPath string) (string, error) {
+	filePath := this.getHashPath(formulaPath)
+
+	hash, err := ioutil.ReadFile(filePath)
+	if err != nil {
+		return "", err
+	}
```
victor-schumacher, line 151: there is a package called `` stream`` with functions for file and dir operations, it might be a good idea to pass it in the constructor, it is much easier to test

pkg/formula/runner/local/pre_run_build.go
```diff
@@ -0,0 +1,202 @@
...
+
+	_ = os.Mkdir(this.getHashDir(), os.ModePerm)
+
+	err = ioutil.WriteFile(filePath, []byte(currentHash), os.ModePerm)
+	if err != nil {
+		return "", err
+	}
```
victor-schumacher, line 168: could you simplifie the error check here?
and use ```stream.FileWriter```

pkg/formula/runner/local/pre_run_build.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+func (this PreRunBuilder) updateHash(formulaPath string) (hash string, err error) {
+	filePath := this.getHashPath(formulaPath)
+	currentHash, err := dirhash.HashDir(formulaPath, "", dirhash.DefaultHash)
+	if err != nil {
+		return "", err
+	}
+
+	_ = os.Mkdir(this.getHashDir(), os.ModePerm)
```
victor-schumacher, line 163: ```stream.DirCreater``` can be a useful here.
why ignore the error?

pkg/formula/runner/local/pre_run_build.go
```diff
@@ -0,0 +1,202 @@
...
+	err = ioutil.WriteFile(filePath, []byte(currentHash), os.ModePerm)
+	if err != nil {
+		return "", err
+	}
+
+	return currentHash, nil
+}
+
+func (this PreRunBuilder) getHashDir() string {
+	return path.Join(this.ritchieHome, hashesPath)
```
victor-schumacher, line 174: what you think about use filepath.Join? i think ```path.Join``` don't work on windows because of the backslashes

victor-schumacher (changes requested):
thanks for contributing!
congrats for your work :1st_place_medal: 
keep going

victor-schumacher:
it stills a draft?

victor-schumacher:
> You can already use the feature in the branch, it was opened as a Draft because:
> 
> * ~Code will be refactored: I'll extract the building part of the pre_run.go to a new file~ **DONE!**
> * Tests: I still need to create the automated tests for this feature
> 
> In parallel, I'd like to get help from the Ritchie team to make some decisions:
> 
> * Currently the modifications are detected by the directory checksum being changed.. I'm saving the checksum of each formula in a separate file under ~/.rit/hashes
>   Do you think this is a great idea or would it be better to save all of them in a single JSON file like ~/.rit/repos/hashes.json ?
>   I don't think marshall/unmarshall times would be that big even for users with a lot of formulas
> * The message is currently:
>   "This formula has changed since the last run, would you like to rebuild?"
>   Is that ok or would you suggest something else?
> * I'm afraid hashing time will be too long for formulas with too many files (like NodeJS formulas with huge node_modules folders), what do you think is the best solution to prevent this? Skip the hash when file count is above a threshold? Limit the hash to a certain directory deepness? (would detect most changes but not all of them) Limit the hashing time? Or is that me overworring about it? Afterall, formulas aren't generally that big?
> 
> Thank you in advance :D

Nice questions @gabriel-pinheiro .

I think it would be nice to store this data directly on formula, but i dont really know if thats possible.
About the json, i don't think it gonna be useful

The message is very good to me.

I don't know, but i think we can let this to the end, so we can test some formulas and see what happen

gabriel-pinheiro:
Thanks for the review, @victor-schumacher !
I'll push the changes soon

> I think it would be nice to store this data directly on formula, but i dont really know if thats possible.
About the json, i don't think it gonna be useful

Yeah, that was my first thought, I couldn't really get it working because if we save the checksum file inside the formula directory, the directory checksum itself would change. That way, the next run would prompt to build again (as there was a change, a file (the checksum) was added/changed).
Other than that, all the formula templates would have to be modified to contain that file in their .gitignore, otherwise git would list it as modified every time :/

> The message is very good to me.

Great!

> I don't know, but i think we can let this to the end, so we can test some formulas and see what happen

Nice, I'll push the requested changes and write some tests so we can move to testing the feature

> it stills a draft?
Yeah, I still want to write some automated tests before converting to a PR

pkg/formula/runner/local/pre_run_build.go
```diff
@@ -0,0 +1,202 @@
...
+	hasExecutable := this.file.Exists(binFilePath)
+	if !hasExecutable {
+		return true
+	}
+
+	// Test: should not check for changes when Bin doesn't exist
+	// Test: should not be blocked by errors
+	// Test: should proceed execution when no changes are made
+	// Test: should proceed execution when user chooses not to rebuild
+	// Test: should rebuild when requested by user
```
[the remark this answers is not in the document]
gabriel-pinheiro: I left them to just remind me of the scenarios I wanted to test, I'll remove them as soon as I commit the tests

pkg/formula/runner/local/pre_run_build.go
gabriel-pinheiro: I wanted to keep both errors because even when getPreviousHash fails I want updateHash to execute. That way, when the checksum file doesn't exist, updateHash will still run creating it. If I returned right after getPreviousHash, updateHash would run.

I could add a comment to make it clearer, what do you think?

pkg/formula/runner/local/pre_run_build.go
gabriel-pinheiro: If it fails because the directory already exists, we still want to continue creating the file

If it fails due to something else, the directory will not be created and the Write below it will fail (and that error is treated)

Is that ok or do you suggest another approach?

pkg/formula/runner/local/pre_run_build.go
victor-schumacher: makes total sense. i think it's a good idea, let's keep the comment on 131, only removing the ```Test:```

pkg/formula/runner/local/pre_run_build.go
victor-schumacher: no problem to me

pkg/formula/runner/local/pre_run_build.go
```diff
@@ -0,0 +1,206 @@
...
+	if err != nil {
+		return false
+	}
+	return ans
+}
+
+func (prb PreRunBuilder) hasFormulaChanged(formulaPath string) (bool, error) {
+	// Test: should always run updateHash, even when getPreviousHash fails
+	previousHash, getErr := prb.getPreviousHash(formulaPath)
+	currentHash, setErr := prb.updateHash(formulaPath)
```
henriquemoraes8, line 133: I wonder if we just compare the hashes first and only update if they are different we could save some I/O

henriquemoraeszup, line 133: I wonder if we just compare the hashes first and only update if they are different we could save some I/O

henriquemoraeszup (commented):
Nice job, just curious, any specific pros and cons you see in using a directory with n hash files instead of a single file with n hash entries?

pkg/formula/runner/local/pre_run_build.go
gabriel-pinheiro: Great idea! Pushing a commit to change it now

gabriel-pinheiro:
> Nice job, just curious, any specific pros and cons you see in using a directory with n hash files instead of a single file with n hash entries?

When I initially thought about this, saving all the hashes to a json file seemed like a costly approach as you would have to read and parse all of them to use only one value in the end but thinking better about it... No one is really gonna have enough formulas to make this process slow.

Other things that also came to mind were:
- Deleting repositories would be much easier, to remove the hashes of the repo xpto we'd only need to remove the `~/.rit/hashes/-home-user-.rit-xpto-*` files
- Updating a hash is much cheaper too, in the JSON approad we would need to save the file with all hashes instead of the hash of one formula

Those wouldn't make that big of a difference for the expected formula amount.. Tbh I don't know if it was worth it to use this approach instead of using the (more organized) single file one.

Do you or anyone else have more points to bring about this? I really have no preferred way tbh haha

gabriel-pinheiro:
By using this version for myself I could notice some improvement points, I'll push them later:
- Building the formula from somewhere else (even the rit build formula) won't update the hash so it'll still prompt you to rebuild. (The fix isn't complicated though, changing it to prompt for rebuild when source changed and bins didn't might be enough, we'll now save two hashes)
- For formulas in the local repository, changes in the workspace (like `~/ritchie-formulas-local/foo/bar`) don't trigger the rebuild as it checks the hash of the repository (`~/.rit/repos/local/foo/bar`) only.

henriquemoraes8:
Hi Gabriel, 

> By using this version for myself I could notice some improvement points, I'll push them later:
> 
> * Building the formula from somewhere else (even the rit build formula) won't update the hash so it'll still prompt you to rebuild. (The fix isn't complicated though, changing it to prompt for rebuild when source changed and bins didn't might be enough, we'll now save two hashes)
> * For formulas in the local repository, changes in the workspace (like `~/ritchie-formulas-local/foo/bar`) don't trigger the rebuild as it checks the hash of the repository (`~/.rit/repos/local/foo/bar`) only.

One question does this only apply to the local repo or are all hashes being compared against the ritchie home? Because the intended use is for the user to **not** edit files in the `.rit` directory. The changes should be detected on the workspace path. Let's say I want to run a formula from my `Zup` workspace (`/Users/henriquerusca/Documents/zup/ritchie-formulas-zup`) the editions on this path should trigger the prompt

henriquemoraes8:
> > Nice job, just curious, any specific pros and cons you see in using a directory with n hash files instead of a single file with n hash entries?
> 
> When I initially thought about this, saving all the hashes to a json file seemed like a costly approach as you would have to read and parse all of them to use only one value in the end but thinking better about it... No one is really gonna have enough formulas to make this process slow.
> 
> Other things that also came to mind were:
> 
> * Deleting repositories would be much easier, to remove the hashes of the repo xpto we'd only need to remove the `~/.rit/hashes/-home-user-.rit-xpto-*` files
> * Updating a hash is much cheaper too, in the JSON approad we would need to save the file with all hashes instead of the hash of one formula
> 
> Those wouldn't make that big of a difference for the expected formula amount.. Tbh I don't know if it was worth it to use this approach instead of using the (more organized) single file one.
> 
> Do you or anyone else have more points to bring about this? I really have no preferred way tbh haha

Yeah, no strong opinion here, your approach definitely scales, but it never might come to this. The tradeoff here is organisation <> maintenance complexity, thinking about it now, I like the way it is easier to simply delete/ovewrite the whole file, so I am ok with this approach 😄

gabriel-pinheiro:
> One question does this only apply to the local repo or are all hashes being compared against the ritchie home? Because the intended use is for the user to **not** edit files in the `.rit` directory. The changes should be detected on the workspace path. Let's say I want to run a formula from my `Zup` workspace (`/Users/henriquerusca/Documents/zup/ritchie-formulas-zup`) the editions on this path should trigger the prompt

Yeah I didn't realize this in the first implementation. Now I'm thinking about implementing it for all workspaces, if the formula has changed in one of them, it will prompt to rebuild.

> Yeah, no strong opinion here, your approach definitely scales, but it never might come to this. The tradeoff here is organisation <> maintenance complexity, thinking about it now, I like the way it is easier to simply delete/ovewrite the whole file, so I am ok with this approach

Nice, I'll have to change the structure a bit for the workspace adaptation so I'll also implement this

henriquemoraeszup:
Hey, this is looking good, just another thing that I remembered to keep in mind. I will try to make the folder structure clear to you:
Currently we can clone a repo or add a workspace to index formulas from anywhere in your file system.
Ritchie will copy the folder structure to the `.rit` path
When you build the formula, it generates a `/bin` folder on the formula directory, not sure if you are hashing just the `src` directly but it could be a good idea since the bin might mess up your hash 😅 

Cheers!

gabriel-pinheiro:
> Hey, this is looking good, just another thing that I remembered to keep in mind. I will try to make the folder structure clear to you:
> Currently we can clone a repo or add a workspace to index formulas from anywhere in your file system.
> Ritchie will copy the folder structure to the `.rit` path
> When you build the formula, it generates a `/bin` folder on the formula directory, not sure if you are hashing just the `src` directly but it could be a good idea since the bin might mess up your hash 
> 
> Cheers!

Great idea! Yeah, I was hashing everything. So, that means all formula templates keep the code inside `src` right? (btw sorry about the delay, I'll probably be able to start coding the changes in a bit)

gabriel-pinheiro:
There we go! Now it's looking for changes in the formula on any workspace and only in the `src` folder. This means that changes in the config.json won't trigger the rebuild, is that ok or would going back to the root of the formula preferred?

**Please don't approve this PR yet** as there are three changes I still want to make:
- For some reason the workspace.List() doesn't return the Default workspace so changes there aren't detected yet, I'll fix it later (probably during lunch today)
- I still need to add unit tests to the new implementation
- Change the hashing structure from the "one file per formula" to the .json

PS: I'm using this version myself to create a formula and the prompt to rebuild is sooo good hahaha. No more need to wait for the `rit build formula --watch` to succeed before running the command

henriquemoraeszup:
Gabriel, no need to fix the list workspace option. It will be fixed on #485 
Nice, can't wait to use it either! hahaha

pkg/formula/runner/pre_run_builder.go
```diff
@@ -0,0 +1,198 @@
...
+		ritchieHome: ritchieHome,
+		workspace:   workspace,
+		builder:     builder,
+		dir:         dir,
+		file:        file,
+		inBool:      inBool,
+	}
+}
+
+func (b PreRunBuilderManager) Build(formulaRelativePath string) {
```
kaduartur, line 73: could you rename the parameter to formulaPath or relativePath?
Go style recommends short variable names

pkg/formula/runner/pre_run_builder.go
```diff
@@ -0,0 +1,198 @@
...
+
+	// First time hashing this formula
+	if err != nil {
+		return false, nil
+	}
+
+	return previousHash != currentHash, nil
+}
+
+func (b PreRunBuilderManager) getPreviousHash(formulaPath string) (string, error) {
```
kaduartur, line 143: Good practice in go is not to add get in the function name. [Getters in golang](https://golang.org/doc/effective_go.html#Getters)

pkg/formula/runner/pre_run_builder.go
```diff
@@ -0,0 +1,198 @@
...
+
+	hash, err := b.file.Read(filePath)
+	if err != nil {
+		return "", err
+	}
+
+	return string(hash), nil
+}
+
+func (b PreRunBuilderManager) getCurrentHash(formulaPath string) (string, error) {
```
kaduartur, line 154: Good practice in go is not to add get in the function name. [Getters in golang](https://golang.org/doc/effective_go.html#Getters)

gabriel-pinheiro:
> Gabriel, no need to fix the list workspace option. It will be fixed on #485

Oh nice!

> As for the config.json it is not clear for me as well, how would you work the root of the formula given the `bin` folder? 2 hashes? One for the `config.json` and one for the `src`?

I'm not comparing the hash of the workspace to the hash of the repo on ~/.rit, I'm comparing the hash of the workspace to its previous hash. That way, as the bin folder isn't generally in the workspace (unless the user runs the makefile in there instead of letting Ritchie build it), it wouldn't mess the hash

pkg/formula/runner/pre_run_builder.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+func (b PreRunBuilderManager) updateHash(formulaPath string, hash string) error {
+	filePath := b.getHashPath(formulaPath)
+
+	_ = b.dir.Create(b.getHashDir())
+	return b.file.Write(filePath, []byte(hash))
+}
+
+func (b PreRunBuilderManager) getHashDir() string {
```
kaduartur, line 165: Do you think that makes sense? it is only used in one place.

pkg/formula/runner/pre_run_builder.go
```diff
@@ -0,0 +1,198 @@
...
+	return filepath.Join(b.ritchieHome, hashesPath)
+}
+
+func (b PreRunBuilderManager) getHashPath(formulaPath string) string {
+	divider := "/"
```
kaduartur, line 170: could you use the `os.PathSeparator` to solve this problem?

pkg/formula/runner/pre_run_builder.go
```diff
@@ -0,0 +1,198 @@
...
+
+	_ = b.dir.Create(b.getHashDir())
+	return b.file.Write(filePath, []byte(hash))
+}
+
+func (b PreRunBuilderManager) getHashDir() string {
+	return filepath.Join(b.ritchieHome, hashesPath)
+}
+
+func (b PreRunBuilderManager) getHashPath(formulaPath string) string {
```
kaduartur, line 169: Good practice in go is not to add get in the function name. [Getters in golang](https://golang.org/doc/effective_go.html#Getters)

pkg/formula/runner/executor.go
```diff
@@ -47,6 +49,7 @@ func (ex ExecutorManager) Execute(exe formula.ExecuteData) error {
 		runner = ex.runners[configType]
 	}
 
+	ex.preRunBuilder.Build(exe.Def.Path)
```
kaduartur, line 52: I think that you should only execute pre-build when the `exe.Def.RepoName` is **local** because other repositories could not be edited. They reflect the git repositories released!

kaduartur (changes requested):
Good job 🚀 If you need to talk about it, just call me!

pkg/formula/runner/pre_run_builder.go
gabriel-pinheiro: Muuch simpler, I didn't know about that one haha! Thanks

pkg/formula/runner/executor.go
gabriel-pinheiro: Ohhh great point

When a formula exists locally, `exe.Def.RepoName` will be `local` even if the formula also exists in a remote repo, right? I've done some tests here and it looks like the case but I just want to make sure, maybe it has something to do with priorities.

I'm asking because someone might be developing locally a formula for a repo and the rebuild prompt wouldn't appear.

I'm pushing this change, if we think about something else I can drop the commit

kaduartur: If you build a formula from a workspace this formula will be built to the local repo. PS, the local repo has the maximum priority.

henriquemoraeszup (commented):
I think I am good to approve, are we missing something here?

gabriel-pinheiro:
I still want to add some changes like the .json for the hashes and invalidate the hash on manual builds but I didn't have enough time this week :(

May we wait until Monday?

gabriel-pinheiro:
Changes made, ready for review again :smiley:

victorschumacherzup (dismissed):

henriquemoraeszup (approved):
giphy (6)

pkg/cmd/mocks_test.go
```diff
@@ -501,6 +501,18 @@ func (w WorkspaceAddListValidatorCustomMock) Validate(workspace formula.Workspac
...
+}
+
+func (w WorkspaceAddListValidatorCustomMock) PreviousHash(string) (string, error) {
+	return "", nil
+}
+
+func (w WorkspaceAddListValidatorCustomMock) UpdateHash(string, string) error {
+	return nil
+}
+
```
brunasilvazup, line 515: Hi @gabriel-pinheiro, thanks for contribuition.

When we use `CustomMock`we have the default to make the internal functions customizable.
With the current change, half of the functions would be customized and half not, which can cause confusion in use.

Could you change that part leaving it customizable and also change the name of the mock to what you redefined `WorkspaceAddListValidatorCustomMock` -> `WorkspaceAddListValidateHasherCustomMock`?

henriquemoraeszup:
@gabriel-pinheiro yours is next to merge, sorry but #485 introduced conflicts 😅

pkg/cmd/mocks_test.go
gabriel-pinheiro: Ohh got it! Pushing a commit now to fix it =P

gabriel-pinheiro:
@brunasilvazup I just pushed the implementation of your suggestions :smile: 

@henriquemoraeszup conflicts resolved =P

pkg/formula/runner/pre_run_builder_test.go
```diff
@@ -0,0 +1,251 @@
...
+				test.previousHashError, test.writeHashError}, builderMock, inputBoolMock)
+			preRunBuilder.Build("/testing/formula")
+
+			if builderMock.HasBuilt() != test.mustBuild {
+				if test.mustBuild {
+					t.Error("Expected formula to build but it didn't")
+				} else {
+					t.Error("Expected formula not to build but it did")
+				}
+			}
```
brunasilvazup, line 181: As always an error is returned, we can remove the if and let `test.mustBuild` itself be the answer something like:

```GO
gotBuild := builderMock.HasBuilt()
if gotBuild != test.mustBuild {
	t.Errorf("Got builder %v, want builder %v", gotBuild, test.mustBuild)
}
```


*I put the build result in a variable because it would be used more than once.

** In the ritchie tests we also follow the got - want pattern and always inform the result and the real expected to have more information.

pkg/formula/runner/pre_run_builder_test.go
```diff
@@ -0,0 +1,251 @@
...
+
+			if inputBoolMock.HasBeenCalled() != test.mustPromptRebuild {
+				if test.mustPromptRebuild {
+					t.Error("Expected formula to prompt for rebuild but it didn't")
+				} else {
+					t.Error("Expected formula not to prompt for rebuild but it did")
+				}
+			}
```
brunasilvazup, line 189: The same logic could be applied here ...

pkg/formula/workspace/workspace.go
```diff
@@ -115,3 +134,40 @@ func (m Manager) List() (formula.Workspaces, error) {
...
 }
+
+func (m Manager) Validate(workspace formula.Workspace) error {
+	dir := workspace.Dir
+	if !m.file.Exists(dir) {
+		return ErrInvalidWorkspace
+	}
+
+	return nil
+}
```
brunasilvazup, line 145: This function was removed in the last PR that entered the master, did you keep it because you will use it specifically or was it an error in the merge?

pkg/formula/runner/pre_run_builder.go
```diff
@@ -0,0 +1,145 @@
...
+		updateErr := b.workspace.UpdateHash(path, currentHash)
+		if updateErr != nil {
+			return false, updateErr
+		}
+	}
+
+	// First time hashing this formula
+	if err != nil {
+		return false, nil
+	}
```
brunasilvazup, line 121: If this error (l 119) refers to the error returned from `PreviousHash` should it not be on the line below it (l 111)?

The same condition is used to enter the previous if (l 111), it made me a little confused...

pkg/formula/workspace/workspace_test.go
```diff
@@ -358,6 +364,184 @@ func TestValidate(t *testing.T) {
...
+
+			if err == nil && tt.out.err != nil {
+				t.Errorf("Validate(%s) expected '%v' error but didn't get any", tt.name, tt.out.err)
+			}
+			if err != nil && tt.out.err == nil {
+				t.Errorf("Validate(%s) didn't expect error but got '%v'", tt.name, err)
+			}
+			if err != nil && tt.out.err != nil && err.Error() != tt.out.err.Error() {
+				t.Errorf("Validate(%s) got error '%v', expected error '%v'", tt.name, err, tt.out.err)
+			}
```
brunasilvazup, line 429: You can simplify these ifs into one with:
```go
if (err != nil) != (tt.out.err != nil) || (err != nil && err.Error() != tt.out.err.Error()) {
	t.Errorf("Validate(%s) got error '%v', expected error '%v'", tt.name, err, tt.out.err)
}
```
The first part validates whether the expected error (whether nil or not) is different from the one received.
`(err != nil) != (tt.out.err != nil)`

The second part is valid if the error is different from null (`err != nil &&`) if the expected and presented errors are the same.
`(err != nil && err.Error() != tt.out.err.Error())`

pkg/formula/workspace/workspace_test.go
```diff
@@ -358,6 +364,184 @@ func TestValidate(t *testing.T) {
...
+			workspace := New(ritHome, ritHome, dirManager, fileManager)
+			err := workspace.UpdateHash(tt.in.formulaPath, tt.in.hash)
+
+			if hashPath != tt.out.path {
+				t.Errorf("Validate(%s) expected hash to be written to %s instead of %s", tt.name, tt.out.path, hashPath)
+			}
+
+			if string(hashContent) != string(tt.out.content) {
+				t.Errorf("Validate(%s) expected hash %s to be written instead of %s", tt.name, string(tt.out.content), string(hashContent))
```
brunasilvazup, line 529: Useful information:

When we use the `tt.Run` structure it already shows the name of the test when presenting the errors, so we don't need to repeat this information in `t.fail` or `Errorf`:

Captura de tela de 2020-10-20 15-20-07

pkg/formula/workspace/workspace_test.go
```diff
@@ -358,6 +364,184 @@ func TestValidate(t *testing.T) {
...
+
+			if err == nil && tt.out.err != nil {
+				t.Errorf("Validate(%s) expected '%v' error but didn't get any", tt.name, tt.out.err)
+			}
+			if err != nil && tt.out.err == nil {
+				t.Errorf("Validate(%s) didn't expect error but got '%v'", tt.name, err)
+			}
+			if err != nil && tt.out.err != nil && err.Error() != tt.out.err.Error() {
+				t.Errorf("Validate(%s) got error '%v', expected error '%v'", tt.name, err, tt.out.err)
+			}
```
brunasilvazup, line 540: If you decide to change, change here too.

brunasilvazup (changes requested):
I left some comments on the testing part, a merge conflict that I think it was and also as a reminder to change the path we talked about.

pkg/formula/workspace/workspace_test.go
gabriel-pinheiro: Ohh elegant solution! Much simpler

pkg/formula/runner/pre_run_builder.go
gabriel-pinheiro: Its down there because I want to update the hash (l 112) when reading the previous hash fails (probably because it's the first time hashing a formula).

The if on 111 is for updating the hash when it's the first time hashing a formula (no previous hash, err != nil) or when the previous hash is different from the current one.

Can you think of a better way to convey this idea in the code structure?

brunasilvazup: I understand now.

And about logic, I think it could be improved by isolating the responsibilities.
A function that says whether or not the hash exists for the formula, if not, makes a return and ends.

Then, another function that is `PreviousHash` and then we only check if there was an error (with normal error return - same as line 106), if it is not verified if the hash has changed following the flow.

What do you think?

brunasilvazup: I understand now

brunasilvazup (approved):
Congratulations on your implementation, ready for QA.

gabriel-pinheiro:
Yayy thank you all so much :smile: 

Btw, just added the .gif of the implementation to the PR, here it is too:
rit2

brunasilvazup:
\merge qa

henriquemoraeszup (approved):

GuillaumeFalourd:
Thank you for contributing @gabriel-pinheiro 👍🏼 

Remembering that if you wish to participate the [Zup challenge for the hacktoberfest](https://insights.zup.com.br/hacktoberfest), you'll have to get 2 PR accepted on Zup repositories. Is this your first contribution, or have you submit something to other repositories as well ?

kaduartur (approved):
LGTM!

gabriel-pinheiro:
Niiice! Can't wait for its version to be release haha

@GuillaumeFalourd it's my first contribution for Zup's Hacktoberfest, I still want to find another Issue to contribute before the end of the month

PR #583, opened 2020-10-03
Set tutorial output more informative

### LINKED ISSUE #525: 'rit tutorial' command shows the same output for enabled or disabled

Different text or color according to my choice

**Why is this needed**:
It is confusing.

Harirai:
Signed-off-by: Harirai <harim1709@gmail.com>
Fixes #525

**- What I did**
Added user response to `rit tutorial` for more clarity, previously output was "Set tutorial successful" irrespective of user response.
Now it shows output corresponding to the input.
Like: "Set tutorial enabled successful".
**- How to verify it**
By setting tutorial.
Set tutorial output more informative.

### DIFF
--- a/pkg/cmd/tutorial.go
+++ b/pkg/cmd/tutorial.go
@@ -68,7 +68,7 @@ func (o tutorialCmd) runStdin() CommandRunnerFunc {
 			return err
 		}
 
-		prompt.Success("Set tutorial " + obj.Tutorial + " successful!")
+		prompt.Success("Tutorial " + obj.Tutorial + "!")
 
 		return nil
 	}
@@ -96,7 +96,7 @@ func (o tutorialCmd) runPrompt() CommandRunnerFunc {
 		if err != nil {
 			return err
 		}
-		prompt.Success("Set tutorial " + response + " successful!")
+		prompt.Success("Tutorial " + response + "!")
 		return nil
 	}
 }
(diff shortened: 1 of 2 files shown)


pkg/cmd/tutorial.go
```diff
@@ -68,7 +68,7 @@ func (o tutorialCmd) runStdin() CommandRunnerFunc {
 			return err
 		}
 
-		prompt.Success("Set tutorial successful!")
+		prompt.Success("Set tutorial " + obj.Tutorial + " successful!")
```
victor-schumacher, line 71: what you think about 
``` go
prompt.Success("Tutorial " + response + "d")
```
for me it's a clearer message

victor-schumacher:
i see there are functional tests failing
to fix this update the file bellow with the new tutorial text:
https://github.com/ZupIT/ritchie-cli/blob/f1b539e9b25bec847e5b71ee97b86f2084175003/functional/core/core_feature.json#L355

pkg/cmd/tutorial.go
Harirai: Yeah sure! 
I assume you want message to be like "Tutorial enabled" , response already contains "d".
So I think it should be    `prompt.Success("Tutorial " + response)` . 
Tell me if I understood it wrong otherwise I'll make the above changes and fix the test soon. :smiley:

victor-schumacher: that's it @Harirai 
keep going

victor-schumacher (approved):

henriquemoraeszup (approved):

PR #598, opened 2020-10-06
[Fix] - Searching only for run.sh on docker pre run

JoaoDanielRufino:
Signed-off-by: JoaoDanielRufino <joaodaniel0405@gmail.com>

**- What I did**
When building a formula with docker on windows, the CLI only checks if there is a run.bat file, but on docker it is necessary to check if there is a run.sh file to build the formula.

**- How to verify it**
Build a formula locally on windows and then run the formula with docker.

Check if there is run.sh file when building formula with docker.

### DIFF
--- a/pkg/formula/formula.go
+++ b/pkg/formula/formula.go
@@ -148,6 +148,10 @@ func (d *Definition) TmpWorkDirPath(home string) string {
 	return filepath.Join(home, TmpDir, u)
 }
 
+func (d *Definition) UnixBinFilePath(fPath string) string {
+	return filepath.Join(fPath, BinDir, BinUnix)
+}
+
 // BinFilePath builds the bin file path from formula path
 func (d *Definition) BinFilePath(fPath string) string {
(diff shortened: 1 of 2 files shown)


GuillaumeFalourd (approved):

lucasdittrichzup (approved):

brunasilvazup:
@JoaoDanielRufino this change its necessary only in windows? If yes, would it be good to specify windows situation in the code?

brunasilvazup (changes requested):
@JoaoDanielRufino this change its necessary only in windows? If yes, would it be good to specify windows situation in the code?

brunasilvazup:
> @JoaoDanielRufino this change its necessary only in windows? If yes, would it be good to specify windows situation in the code?
> 
> PS.: I used 'request changes' to discuss this before, not necessarily needs changes.

We talked and I could understand that it is just a check for windows.

kaduartur (approved):

brunasilvazup (approved):

PR #601, opened 2020-10-07
Fix timer os metric build formula watch

brunats:
**- What I did**
Add stop condition to the hit build formula --watch command. Now the process can be completed correctly when the user presses 'Ctrl + c'.

This corrects the timing of sending metrics that is next to the process stop.

>stop

**- How to verify it**

Fix timer os metric build formula watch


pkg/formula/watcher/watcher.go
```diff
@@ -78,11 +97,9 @@ func (w *WatchManager) Watch(workspacePath, formulaPath string) {
 
 	w.build(workspacePath, formulaPath)
 
-	watchText := fmt.Sprintf("Watching dir %s \n", formulaPath)
-	prompt.Info(watchText)
-
-	startTime := time.Now().Second()
-	w.sendMetric(float64(startTime))
```
henriquemoraeszup, line 85: just curious, where is the start time being collected now? And how does it differ a start from an end? `sendMetric` interprets it on the 2nd call?

brunasilvazup: Hi, I also had this question, I just changed the location, but I didn't check the difference.

In the main it uses a function to calculate the difference.
And checking now I saw that leaving the send metric here he sends twice.

In the case of the main send metric it uses save the start time and get the time again at the end of the execution.

pkg/formula/watcher/watcher.go
```diff
@@ -54,8 +57,22 @@ func New(
 	}
 }
 
+func (w *WatchManager) closeWatch() {
+	currentTime := time.Now().Second()
+	w.sendMetric(float64(currentTime))
```
henriquemoraeszup, line 62: is this asynchronous or should the `stopping` text come before it?

henriquemoraeszup (approved):
Just one comment but lgtm

pkg/formula/watcher/watcher.go
brunasilvazup: sendMetric is asynchronous (:

victorschumacherzup (approved):

brunasilvazup:
\merge qa

lucasdittrichzup (approved):

Issue #594, opened 2020-10-05
Update all formulas repositories at once


GuillaumeFalourd:
**What would you like to be added**:

The possibility to update all formulas repositories together.

**Why is this needed**:

Today, the `rit update repo` command only update one formula repository at a time. It would be interesting to have the possibility to update all repositories with one command:

```
➜  ~ rit update repo
? Select a repository to update:   [Use arrows to move, type to filter]
> ALL
  commons
  demo
  formulas-insights
```

And then select which version I want to use for each repo during the execution.

Harirai:
Hii @GuillaumeFalourd, I'd like to work on this issue. 
Can you guide me and help me get started?
I'm new to the Go language, going through its documentation and trying to understand file: `pkg/cmd/update_repo.go`.
Am I going in the right direction?

GuillaumeFalourd:
Hi @Harirai

Yes, you're going in the right direction 👍 

My suggestion would be for you to test the rit update repo command first, to understand how it works.
Then, think of a solution to execute it only once for all repositories contained in the list.
Moreover, it may not be necessary to ask the user to update a repo that is already with the last available version.

Start with this, and let us know if you have any question.

PR #602, opened 2020-10-08
Added update All for repo

Harirai:
### -  What I did
Added an option "ALL" for `rit update repo`, which allows users to update all the repositories in one go.
Fixes issue #594 
### - How to verify it
By running it locally after changes.
ritUpdate

### - Description for changelog
Added update ALL for repo.


Harirai:
Hi @GuillaumeFalourd, What do you think about current changes?

GuillaumeFalourd:
A few suggestions @Harirai 

- You don't need to comment the code, we encourage to name variables and methods to be self explanatory.
- If you use a string more than once on a class, create a const variable instead of duplicating the string.

I'll add Ritchie Core Team reviewers for them to help you implementing the feature following the right structure.
@kaduprussekzup @victorschumacherzup @brunasilvazup

Harirai:
I have checked the update for ALL locally, 
Screenshot from 2020-10-13 18-38-35

Currently, I don't know how to update for "Stdin".

GuillaumeFalourd:
> Currently, I don't know how to update for "Stdin".

It may not be necessary to make this option `ALL` available for the STDIN command. 

As the STDIN has been created to automate processes (calling formulas automatically inside an implementation), the user could use it directly for specific repositories if needed. What do you think ?

pkg/cmd/update_repo.go
```diff
@@ -97,33 +97,39 @@ func (up updateRepoCmd) runPrompt() CommandRunnerFunc {
...
+
+		var repoToUpdate []formula.Repo
 		for i := range repos {
-			if repos[i].Name == formula.RepoName(name) {
-				repo = repos[i]
-				break
+			if repos[i].Name == formula.RepoName(name) || flagAll {
+				repoToUpdate = append(repoToUpdate, repos[i])
 			}
 		}
```
brunasilvazup, line 107: Hi @Harirai,
Thinking about the logic of ` flagAll`, when it is selected the resulting list (`repoToUpdate`) will be equal to `repos`.

In this case we don't need the for, so a check could be done before the for in the case of ALL.
So, the code flow would not need to pass through the 'for' flow.

What do you think?

pkg/cmd/update_repo_test.go
```diff
@@ -31,6 +31,10 @@ import (
 	"github.com/ZupIT/ritchie-cli/pkg/prompt"
 )
 
+const (
```
henriquemoraeszup, line 34: we should have a test for the `ALL` option

victorschumacherzup:
Hi @Harirai.
Thanks for the contribution!
it is ready for review and merge?

pkg/cmd/update_repo.go
Harirai: Hi @brunasilvazup, that's a great suggestion. I'll implement it accordingly, Thank you!

pkg/cmd/update_repo.go
```diff
@@ -97,33 +97,39 @@ func (up updateRepoCmd) runPrompt() CommandRunnerFunc {
...
+			repoInfo := git.NewRepoInfo(currRepo.Url, currRepo.Token)
+			tags, err := git.Repos.Tags(repoInfo)
+			if err != nil {
+				return err
+			}
 
-		if err := up.repo.Update(formula.RepoName(name), formula.RepoVersion(version)); err != nil {
-			return err
-		}
+			var currRepoVersion string = fmt.Sprintf("Select your new version for %q:", currRepo.Name.String())
```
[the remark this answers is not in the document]
Harirai: Yeah sure!

pkg/cmd/update_repo.go
```diff
@@ -88,6 +87,7 @@ func (up updateRepoCmd) runPrompt() CommandRunnerFunc {
 		}
 
 		var reposName []string
+		reposName = append(reposName, "ALL")
```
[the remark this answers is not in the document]
Harirai: Yes that will be a better, thank you!

pkg/cmd/update_repo_test.go
Harirai: Yes, I was thinking about that, but as I'm new to testing and Go language, I think It will take some time for me.
Maybe I (or anyone else) can solve it by opening another issue. Will it be okay?

brunasilvazup:
Hi @Harirai, thanks for your contribuition and remember of adds your name in `CONTRIBUTORS` file (:

pkg/cmd/update_repo_test.go
GuillaumeFalourd: I personally think this issue would be a good opportunity for you to learn how to perform unit test in Golang @Harirai 
Nothing is urgent, you can take your time to understand how the class has been implemented and how you could add the scenario related to your implementation.

pkg/cmd/update_repo.go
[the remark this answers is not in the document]
victorschumacherzup: No problem @Harirai, two ways to declare a variable can be confusing.

Quoting from [stackoverflow](https://stackoverflow.com/questions/27919359/why-there-are-two-ways-of-declaring-variables-in-go-whats-the-difference-and-w)
> One of Go's guiding design principle was to make the syntax clean. Many statements require or it is handy that they allow declaring local variables which will be only available in the body of the statement such as for, if, switch etc. To make the syntax cleaner and shorter, Short variable declaration is justified in these cases and it is unambigous what they do.

All the other inside function variables are short declared, don't make sense to have only different one right?

GuillaumeFalourd:
Hi @Harirai , did you conclude your implementation ? 😄

Harirai:
Hi @GuillaumeFalourd, sorry for the delay :sweat_smile:. Currently, I'm working on some academic projects and assignments. I'll try to complete the tests as soon as possible.
Incase it is urgent, please feel free to assign the issue to anyone else. :smile:

GuillaumeFalourd:
It's not urgent @Harirai , it's just I wasn't sure if you would be working on the tests. Take your time and good luck with your other assignments 👍🏼

brunasilvazup (approved):

lucasdittrichzup (approved):

lucasdittrichzup:
great job @Harirai :rocket:

Harirai:
@lucasdittrichzup  Thank you, I learnt a lot. Thanks again for always helping. @GuillaumeFalourd

GuillaumeFalourd:
Thank you again @Harirai , and feel free to keep going and use / share Ritchie CLI 🚀 
We are eager to receive more feedback about the tool 😄

PR #612, opened 2020-10-15
Add multiselect option

### LINKED ISSUE #600: Add a multiselect input type for formulas

I would like to see a new input type to allow selecting more than one option for a same input.

**Why is this needed**:

Today, there is only 3 types: `text`, `bool`, `password`, and be can use the `list` field to give options for the user to choose. However, you can only choose one item on the list at the moment, it would be great to be able to select more than one option for some inputs when necessary.

**Implementation Details**

This feature is already available in the lib survey that is used in the cli. Then just implement a new type of content for the config.json inputs called multiselect.

The behavior will be similar to the list type, but using this multiselect feature.

Example:
https://github.com/AlecAivazis/survey#multiselect

image

rafaeelaudibert:
**- What I did**
**[WIP]** Added multiselect option to Ritchie
Closes #600 

**- How to verify it**

Added multiselect option to Ritchie

### DIFF
--- a/pkg/prompt/multiselect.go
+++ b/pkg/prompt/multiselect.go
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2020 ZUP IT SERVICOS EM TECNOLOGIA E INOVACAO SA
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package prompt
+
+import (
+	"github.com/AlecAivazis/survey/v2"
+)
+
+type SurveyMultiSelect struct{}
+
+func NewSurveyMultiSelect() SurveyMultiSelect {
+	return SurveyMultiSelect{}
+}
+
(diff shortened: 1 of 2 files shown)


rafaeelaudibert:
I am having trouble trying to work around the code. I added the basic interface and integration with the survey library, but didn't understand where I should connect this code to, so that I can handle incoming 'multiselect' types of formulas

GuillaumeFalourd:
> I am having trouble trying to work around the code. I added the basic interface and integration with the survey library, but didn't understand where I should connect this code to, so that I can handle incoming 'multiselect' types of formulas

@rafaeelaudibert : as you're adding a new input type, the first place to update should be the inputs.go class : https://github.com/ZupIT/ritchie-cli/blob/master/pkg/formula/runner/inputs.go . Then, you should check where methods from this class are being used and analyze what should be updated (implementation + tests).

henriquemoraeszup:
Just to provide extra context, when you run `rit formula create` you create a formula with a `config.json` file in which you can define the interactive inputs. The file guillaume mentioned is responsible for reading the config during a formula execution and prompting the user with the correct form

rafaeelaudibert:
Oh, I'm sorry, I've clearly gone in the wrong way. Will try to work on it some more (in the right place this time 😄 ) and will come back with an improved PR (in this same one actually, I'll just rebase it). Thanks!

GuillaumeFalourd:
Hi @rafaeelaudibert , how are you doing ?

Are you still working on this ISSUE ?

rafaeelaudibert:
Hi @GuillaumeFalourd, how are you?

Sadly, I won't have the time to finish this issue right now (end of semester at uni).

I might work this in the future, but not in a close one. Feel free to close this, if anybody else is interested in the issue.
Thanks for the help!

GuillaumeFalourd:
No problem @rafaeelaudibert , I believe the Ritchie Core Team will work on it then, and once you'll be available to contribute again, feel free to work on another ISSUE available on our repositories 👍🏼 

Thank you again for your interest about Ritchie 🚀

PR #616, opened 2020-10-15
feat: add default field visualization

DittrichLucas:
**- What I did**

Show the default input value in parentheses at the end of the question.

default

This was necessary to make it more evident to the user that, by not informing value in an input that has a default value, it will be used.

Initially, the types of input that accept this improvement are: text and int. For the other types (bool, email, list, multiline and password) this improvement was not efficient.

**- How to verify it**

Run a formula that has input with default value.

Show the default input value in parentheses at the end of the question.

### DIFF
--- a/pkg/cmd/create_formula.go
+++ b/pkg/cmd/create_formula.go
@@ -277,7 +277,7 @@ func FormulaWorkspaceInput(
 	}
 
 	items = append(items, newWorkspace)
-	selected, err := inList.List("Select a formula workspace: ", items, "")
+	selected, err := inList.List("Select a formula workspace: ", items)
 	if err != nil {
 		return formula.Workspace{}, err
 	}
--- a/pkg/cmd/create_formula.go
+++ b/pkg/cmd/create_formula.go
@@ -103,7 +103,7 @@ func (c createFormulaCmd) runPrompt() CommandRunnerFunc {
 			return err
 		}
 
-		lang, err := c.inList.List("Choose the language: ", languages)
+		lang, err := c.inList.List("Choose the language: ", languages, "")
 		if err != nil {
 			return err
 		}
--- a/pkg/cmd/create_formula.go
+++ b/pkg/cmd/create_formula.go
@@ -103,7 +103,7 @@ func (c createFormulaCmd) runPrompt() CommandRunnerFunc {
 			return err
 		}
 
-		lang, err := c.inList.List("Choose the language: ", languages)
+		lang, err := c.inList.List("Choose the language: ", languages, "")
 		if err != nil {
 			return err
 		}
@@ -277,7 +277,7 @@ func FormulaWorkspaceInput(
 	}
 
 	items = append(items, newWorkspace)
-	selected, err := inList.List("Select a formula workspace: ", items)
+	selected, err := inList.List("Select a formula workspace: ", items, "")
 	if err != nil {
 		return formula.Workspace{}, err
 	}
@@ -286,12 +286,12 @@ func FormulaWorkspaceInput(
 	var workspacePath string
 	var wspace formula.Workspace
 	if selected == newWorkspace {
-		workspaceName, err = inText.Text("Workspace name: ", true)
+		workspaceName, err = inText.Text("Workspace name: ", true, "")
 		if err != nil {
 			return formula.Workspace{}, err
 		}
 
-		workspacePath, err = inText.Text("Workspace path (e.g.: /home/user/github):", true)
+		workspacePath, err = inText.Text("Workspace path (e.g.: /home/user/github):", true, "")
 		if err != nil {
 			return formula.Workspace{}, err
 		}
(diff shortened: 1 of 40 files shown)


henriquemoraeszup (commented):
I am a bit confused, I thought we just needed to display the text of the default value in case the user defined a default value. Maybe inputs like list or bool do not even require this refactor. Imagine you will always display a list of items, if the user presses enter straight away, then the first item is always selected, much like a default value.

What do you think?

pkg/cmd/tutorial.go
```diff
@@ -87,7 +87,7 @@ func (o tutorialCmd) runPrompt() CommandRunnerFunc {
 		tutorialStatusCurrent := tutorialHolder.Current
 		fmt.Println("Current tutorial status: ", tutorialStatusCurrent)
 
-		response, err := o.List(msg, statusTypes)
+		response, err := o.List(msg, statusTypes, statusTypes[0])
```
henriquemoraeszup, line 90: Is this list staying?

pkg/cmd/upgrade.go
```diff
@@ -67,7 +67,7 @@ func (u UpgradeCmd) runFunc() CommandRunnerFunc {
 		if !u.file.Exists(metric.FilePath) {
 
 			options := []string{AcceptMetrics, DoNotAcceptMetrics}
-			choose, err := u.input.List(AddMetricsQuestion, options)
+			choose, err := u.input.List(AddMetricsQuestion, options, options[1])
```
henriquemoraeszup, line 70: This list as well?

pkg/cmd/create_formula.go
```diff
@@ -103,7 +103,7 @@ func (c createFormulaCmd) runPrompt() CommandRunnerFunc {
 			return err
 		}
 
-		lang, err := c.inList.List("Choose the language: ", languages)
+		lang, err := c.inList.List("Choose the language: ", languages, "")
```
henriquemoraeszup, line 106: Are we removing default values from lists?

pkg/cmd/init.go
```diff
@@ -260,7 +260,7 @@ You can view our Privacy Policy (http://insights.zup.com.br/politica-privacidade
 }
 
 func (in initCmd) setRunnerType() error {
-	selected, err := in.List(SelectFormulaTypeQuestion, formula.RunnerTypes)
+	selected, err := in.List(SelectFormulaTypeQuestion, formula.RunnerTypes, formula.RunnerTypes[0])
```
henriquemoraeszup, line 263: How about this list?

henriquemoraeszup (commented):
I see we have some lists still, is this intentional? If we remove them we actually reduce the number of files edited, which is good.

Another thing, if we had a new component like `InputTextDefault` we could avoid refactoring the code in multiple files by changing only the formula one, do you think its valid? I do not have a strong opinion, just to think about it =]

pkg/prompt/int.go
```diff
@@ -31,20 +31,20 @@ func NewSurveyInt() SurveyInt {
 	return SurveyInt{}
 }
 
-func (SurveyInt) Int(name string, helper ...string) (int64, error) {
+func (SurveyInt) Int(name string, defaultValue string, helper ...string) (int64, error) {
```
victorschumacherzup, line 34: is possible to declare the parameter only once here? like: 
``` go 
func Int(name, defaultValue string, helper ...string)
```

pkg/prompt/int.go
```diff
@@ -31,20 +31,20 @@ func NewSurveyInt() SurveyInt {
...
 	var value string
+	var input = &survey.Input{Message: name, Default: defaultValue}
```
victorschumacherzup, line 36: i think a simplified variable declaration would be cleaner ```:=```

pkg/prompt/text.go
```diff
@@ -26,13 +26,15 @@ func NewSurveyText() SurveyText {
 	return SurveyText{}
 }
 
-func (SurveyText) Text(name string, required bool, helper ...string) (string, error) {
+func (SurveyText) Text(name string, required bool, defaultValue string, helper ...string) (string, error) {
 
 	var value string
+	var input = &survey.Input{Message: name, Default: defaultValue}
```
victorschumacherzup, line 32: same observations here

pkg/cmd/create_formula.go
```diff
@@ -286,12 +286,12 @@ func FormulaWorkspaceInput(
...
 	var wspace formula.Workspace
 	if selected == newWorkspace {
-		workspaceName, err = inText.Text("Workspace name: ", true)
+		workspaceName, err = inText.Text("Workspace name: ", true, "")
 		if err != nil {
 			return formula.Workspace{}, err
 		}
 
-		workspacePath, err = inText.Text("Workspace path (e.g.: /home/user/github):", true)
+		workspacePath, err = inText.Text("Workspace path (e.g.: /home/user/github):", true, "")
```
victorschumacherzup, line 294: The third arg ir necessary? I think you can pass only two because ```Text``` receive a ```...string```

lucasdittrichzup:
due to the amount of difference to be passed and the scope changes, I am closing this PR to open a new

PR #632, opened 2020-10-23
added context printing

victorschumacherzup:
### Description

Most of users want to be sure about which context your formula will be ran, a common behaviour about this is to use ```rit set context``` just to see the context.

Printing the context after the start of every formula can help on this 

image

### How to verify it
Run some formula

### Changelog
added context printing

### DIFF
--- a/pkg/formula/runner/local/runner.go
+++ b/pkg/formula/runner/local/runner.go
@@ -102,6 +102,13 @@ func (ru RunManager) setEnvs(cmd *exec.Cmd, pwd string, verbose bool) error {
 		return err
 	}
 
+	ctxToPrint := ctx.Current
+	if ctxToPrint == "" {
+		ctxToPrint = "default"
+	}
+	prompt.Info(
+		fmt.Sprintf("Formula running  on context: %s\n", prompt.Cyan(ctxToPrint)),
+	)
 	cmd.Env = os.Environ()
 	dockerEnv := fmt.Sprintf(formula.EnvPattern, formula.DockerExecutionEnv, "false")
 	pwdEnv := fmt.Sprintf(formula.EnvPattern, formula.PwdEnv, pwd)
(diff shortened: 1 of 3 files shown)


pkg/formula/runner/local/runner.go
```diff
@@ -102,6 +102,13 @@ func (ru RunManager) setEnvs(cmd *exec.Cmd, pwd string, verbose bool) error {
 		return err
 	}
 
+	ctxToPrint := ctx.Current
+	if ctxToPrint == "" {
+		ctxToPrint = "default"
+	}
+	prompt.Info(
+		fmt.Sprintf("Formula running  on context: %s\n", prompt.Cyan(ctxToPrint)),
```
brunasilvazup, line 110: Typo: two spaces between 'running' and 'on'.

lucasdittrichzup (dismissed):

pkg/formula/runner/local/runner.go
```diff
@@ -102,6 +102,13 @@ func (ru RunManager) setEnvs(cmd *exec.Cmd, pwd string, verbose bool) error {
 		return err
 	}
 
+	ctxToPrint := ctx.Current
+	if ctxToPrint == "" {
+		ctxToPrint = "default"
+	}
```
henriquemoraeszup, line 108: How about not printing the context at all if the user only has one default context? This way, new users do not have to process the concept of something they will not use until later on

pkg/formula/runner/local/runner.go
victorschumacherzup: removed, thanks

pkg/formula/runner/local/runner.go
victorschumacherzup: That's make sense, behavior changed on 23ead81

brunasilvazup (approved):

PR #633, opened 2020-10-26
added warning on same env name

### LINKED ISSUE #625: Warn machine env variables if they have the same name as input params
Notify the user with a warning if they have an input param with the same name as one of their bash envs

**Why is this needed**:
Imagine the user defines an input param called `path`, bam, he will lose all commands in that executing process

victorschumacherzup:
### Description

Closes #625 

Using the function ```LookupEnv``` to check. Documentation [here](https://golang.org/pkg/os/#LookupEnv)

Output:
image

### Changelog
added warning on same env name

### DIFF
--- a/pkg/formula/runner/inputs_test.go
+++ b/pkg/formula/runner/inputs_test.go
@@ -101,7 +101,7 @@ func TestInputManager_Inputs(t *testing.T) {
 	var inputs []formula.Input
 	_ = json.Unmarshal([]byte(inputJson), &inputs)
 	_ = os.Setenv("SAMPLE_TEXT", "someValue")
-	
+
 	setup := formula.Setup{
 		Config: formula.Config{
 			Inputs: inputs,
--- a/pkg/formula/runner/inputs_test.go
+++ b/pkg/formula/runner/inputs_test.go
@@ -100,7 +100,8 @@ func TestInputManager_Inputs(t *testing.T) {
 ]`
 	var inputs []formula.Input
 	_ = json.Unmarshal([]byte(inputJson), &inputs)
-
+	_ = os.Setenv("SAMPLE_TEXT", "someValue")
+	
 	setup := formula.Setup{
 		Config: formula.Config{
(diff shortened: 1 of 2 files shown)


GuillaumeFalourd:
I'm wondering. This message will be sent only during the formula execution, right ?
I get that the one implementing the formula will be able to check this during his/her tests, but he/she may not have the same local variable set than other users.

If someone execute the formula for the first time without knowing the variables names, shouldn't we send a message a bit more explicit ? Such as `you're about to overwrite the local variable $PATH, do you want to proceed ?` or something like that ?

victorschumacherzup:
> I'm wondering. This message will be sent only during the formula execution, right ?
> I get that the one implementing the formula will be able to check this during his/her tests, but he/she may not have the same local variable set than other users.
> 
> If someone execute the formula for the first time without knowing the variables names, shouldn't we send a message a bit more explicit ? Such as `you're about to overwrite the local variable $PATH, do you want to proceed ?` or something like that ?

formulas don't overwrite local variables, it use the value of them

lucasdittrichzup (approved):

henriquemoraeszup (approved):
I am in favor of not blocking the user with another prompt, because if the variable in question is irrelevant (i.e.: username), we could unnecessarily block the user on every execution

pkg/formula/runner/inputs_test.go
```diff
@@ -100,6 +100,7 @@ func TestInputManager_Inputs(t *testing.T) {
 ]`
 	var inputs []formula.Input
 	_ = json.Unmarshal([]byte(inputJson), &inputs)
+	_ = os.Setenv("SAMPLE_TEXT", "someValue")
```
henriquemoraeszup, line 103: ideally we should check for the warning text on stdout, hopefully @brunasilvazup will figure it out =]

brunasilvazup (approved):

GuillaumeFalourd (approved):

Issue #615, opened 2020-10-15
Add stdin capacity to rit build formula


viniciussousazup:
**What would you like to be added**:
Add flag `--stdin` to `rit build formula` 

**Why is this needed**:
Today the `rit build formula` need user interaction, this is a problem to CI integration and general automation.

rodrigo-brito:
Hi @viniciussousazup, I will try to solve this.

viniciussousazup:
Hi @rodrigo-brito, I am not working on it, If you want you can take it.

rodrigo-brito:
No problem, I will try another one.

GuillaumeFalourd:
> No problem, I will try another one.

@rodrigo-brito , I've assigned the issue to you as @viniciussousazup said he wasn't working on it.

But let me know if you want to work on another one, and I'll remove the assignment 👍🏼

rodrigo-brito:
Ok, I will try this weekend.

aronrichter:
Hey guys... how's going? Can I take this issue?

GuillaumeFalourd:
Hi @aronrichter , thank you for your message.
We didn't get any return from @rodrigo-brito until know so if you're interested in working on this ISSUE, I can assign it to you as well 👍🏼

rodrigo-brito:
Hi @aronrichter and @GuillaumeFalourd. Sorry for the delay in the report. 
I am focused on another issue this week, you can take it. Have fun!

brunasilvazup:
Unfortunately, we have a new PR (https://github.com/ZupIT/ritchie-cli/pull/692) that will disable the `rit build formula` command, which is the main change in this Issue.
Thus, we will close this issue.

PR #648, opened 2020-10-31
add -stdin flag on build formula command

aronrichter:
### Description
As describe on Issue #615:
Today the `rit build formula` need user interaction, this is a problem to CI integration and general automation.

### Changelog
add `--stdin` flag on build formula command

### DIFF
--- a/pkg/cmd/build_formula.go
+++ b/pkg/cmd/build_formula.go
@@ -29,6 +29,7 @@ import (
 	"github.com/ZupIT/ritchie-cli/pkg/prompt"
 	"github.com/ZupIT/ritchie-cli/pkg/rtutorial"
 	"github.com/ZupIT/ritchie-cli/pkg/slice/sliceutil"
+	"github.com/ZupIT/ritchie-cli/pkg/stdin"
 	"github.com/ZupIT/ritchie-cli/pkg/stream"
 )
 
@@ -50,6 +51,11 @@ type buildFormulaCmd struct {
 	prompt.InputText
 	prompt.InputList
 	rt rtutorial.Finder
+}
+
+type buildFormulaStdin struct {
+	WorkspacePath string `json:"workspacePath"`
+	Formula       string `json:"formula"`
 }
 
 func NewBuildFormulaCmd(
@@ -78,7 +84,7 @@ func NewBuildFormulaCmd(
 		Short: "Build your formulas locally. Use --watch flag and get real-time updates.",
 		Long: `Use this command to build your formulas locally. To make formulas development easier, you can run
 the command with the --watch flag and get real-time updates.`,
-		RunE:      s.runFunc(),
+		RunE:      RunFuncE(s.runStdin(), s.runPrompt()),
 		ValidArgs: []string{""},
 		Args:      cobra.OnlyValidArgs,
 	}
@@ -87,7 +93,7 @@ the command with the --watch flag and get real-time updates.`,
 	return cmd
 }
 
-func (b buildFormulaCmd) runFunc() CommandRunnerFunc {
+func (b buildFormulaCmd) runPrompt() CommandRunnerFunc {
 	return func(cmd *cobra.Command, args []string) error {
 		workspaces, err := b.workspace.List()
 		if err != nil {
@@ -129,6 +135,35 @@ func (b buildFormulaCmd) runFunc() CommandRunnerFunc {
 			return err
 		}
 		tutorialBuildFormula(tutorialHolder.Current)
+
+		return nil
+	}
+}
+
+func (b buildFormulaCmd) runStdin() CommandRunnerFunc {
+	return func(cmd *cobra.Command, args []string) error {
+		buildFormulaStdin := buildFormulaStdin{}
+
+		if err := stdin.ReadJson(cmd.InOrStdin(), &buildFormulaStdin); err != nil {
+			return err
+		}
+
+		if err := buildCommandValidator(b, buildFormulaStdin); err != nil {
+			return err
+		}
+
+		formulaPath := filepath.Join(buildFormulaStdin.WorkspacePath, buildFormulaStdin.Formula)
+
+		wspace := formula.Workspace{
+			Name: buildFormulaStdin.Formula,
+			Dir:  formulaPath,
+		}
+
+		info := formula.BuildInfo{
+			FormulaPath: formulaPath,
+			Workspace:   wspace,
+		}
+		b.build(info)
 
 		return nil
 	}
@@ -215,6 +250,22 @@ func hasFormulaInDir(dirs []string) bool {
 	dirs = sliceutil.Remove(dirs, srcDir)
 
 	return len(dirs) > 0
+}
+
+func buildCommandValidator(b buildFormulaCmd, buildFormulaStdin buildFormulaStdin) error {
+	dirs, err := b.directory.List(buildFormulaStdin.WorkspacePath, false)
+
+	if err != nil {
+		return prompt.NewError("dir does not exists")
+	}
+
+	for _, dir := range dirs {
+		if dir == buildFormulaStdin.Formula {
+			return nil
+		}
+	}
+
+	return prompt.NewError("formula does not exists")
 }
 
 func tutorialBuildFormula(tutorialStatus string) {
--- a/pkg/cmd/build_formula_test.go
+++ b/pkg/cmd/build_formula_test.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 
 	"github.com/ZupIT/ritchie-cli/pkg/formula"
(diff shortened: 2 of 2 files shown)


GuillaumeFalourd:
Hi @aronrichter, thank you for contributing 👍🏼 

I'll let the CODEOWNERS check everything is ok when they'll be available, and I'll already add the label for your contribution during the #Hacktoberfest! 🚀

henriquemoraeszup (approved):
Nice job 🚀

pkg/cmd/build_formula.go
```diff
@@ -130,6 +136,30 @@ func (b buildFormulaCmd) runFunc() CommandRunnerFunc {
...
+			return err
+		}
+
+		err = buildCommandValidator(b, buildFormulaStdin)
+
+		if err != nil {
+			return err
+		}
+
+		formulaPath := buildFormulaStdin.WorkspacePath+"/"+buildFormulaStdin.Formula
```
henriquemoraeszup, line 155: You need to use `filepath.Join` to support all OSes

henriquemoraeszup (changes requested):
Just found a small correction for cross-platform stability

pkg/cmd/build_formula.go
victorschumacherzup, line 155: this "/" can be a problem with windows, can you use `filepath.Join` ?

victorschumacherzup (changes requested):

henriquemoraeszup (approved):

victorschumacherzup (approved):

pkg/cmd/build_formula.go
```diff
@@ -130,6 +136,30 @@ func (b buildFormulaCmd) runFunc() CommandRunnerFunc {
 	}
 }
 
+func (b buildFormulaCmd) runStdin() CommandRunnerFunc {
+	return func(cmd *cobra.Command, args []string) error {
+		buildFormulaStdin := buildFormulaStdin{}
+
+		err := stdin.ReadJson(cmd.InOrStdin(), &buildFormulaStdin)
+
+		if err != nil {
```
brunasilvazup, line 145: Do you can join this lines in only one with:
`if err := stdin.ReadJson(cmd.InOrStdin(), &buildFormulaStdin); err != nil {`

what do you think?

pkg/cmd/build_formula.go
```diff
@@ -130,6 +136,30 @@ func (b buildFormulaCmd) runFunc() CommandRunnerFunc {
...
+			return err
+		}
+
+		err = buildCommandValidator(b, buildFormulaStdin)
+
+		if err != nil {
```
brunasilvazup, line 151: Same suggestion here

brunasilvazup (dismissed):

lucasdittrichzup (approved):

pkg/cmd/build_formula.go
```diff
@@ -52,6 +53,11 @@ type buildFormulaCmd struct {
 	rt rtutorial.Finder
 }
 
+type buildFormulaStdin struct {
```
lucasdittrichzup, line 56: @aronrichter, I did some testing with its implementation and I have two considerations:

1. when I pass the empty `--stdin` flag (`rit build formula --stdin`) the process gets stuck

2. it would be interesting to add the flag on the helper of the command, as it will help anyone who wants to use it

lucasdittrichzup (changes requested):

pkg/cmd/build_formula.go
aronrichter: Hey @lucasdittrichzup 

Do you know if this error happen only on this command?
I ask this 'cause I used the same pattern as others commands.
If this problem happen with others commands too, maybe will be a good idea treat this thing on another issue, because will affect every command already developed.

At the mean time, I'll add the flag on helper.

Thank you for the review :grimacing:

brunasilvazup: Hi @aronrichter and @lucasdittrichzup
I checked and this behavior is repeated in other commands too. So do you continue with your current implementation.

I will do other tests in QA, thanks for your contribution (:

brunasilvazup (commented):
Hi @aronrichter,

Can you resolve the latest conflicts, please?

brunasilvazup:
Hi @aronrichter,
I create a new formula (`rit teste build-stdin`) in workspace default to tests and tested your implementation with:

```shell
echo '{"workspacePath":"/home/bruna/ritchie-formulas-local", "formula": "rit teste build-stdin"}' | rit build formula --stdin
```
And this returned an error: `Error: formula does not exists`.

I think there may be a problem when searching for the formula on the `buildCommandValidator` on line 262, as it is always looking for a directory that has the complete formula and the formula will be each part a directory.

aronrichter:
Hi @brunasilvazup ... 

First of all, thank you for your review :grin: 

I was looking at the code, and correct me if I'm wrong, but I understood the rule this way:

When you use the command `rit create formula`, you'll have a dir with the formula's name, and inside of this dir your formula will exist.
The goal of the build formula is take the created formula inside of its dir and just build it;

With that in mind, when you use the prompt option to build a formula, you pass through these steps:
1. select a formula workspace;
2. select a formula or group:
   2.1 select a formula name to Ritchie build it;
   2.2 choose a group, to go down on the file tree, and then select a formula name;

When you select a formula, Ritchie will enter on chosen dir and then build the formula inside using one of those implementations of Build interface:
- local.go
- shell.go 
- docker.go
- make.go
- bat.go

You can see this behavior in `build_formula.go` on line 181 when the program call `if err := b.formula.Build(info); err != nil {`.

When I developed this feature, I tryed to focus on this. 
So, `Workspace` should be the place of your formula's dir, and `Formula` should be the name of your formula, therefore the name of the formula's dir. Keeping the same behavior already implemented on prompt option.

As far I understood of the behavior of Ritchie,  your test is validating a uncreated formula, so would be right you get this error: `Error: formula does not exists`
To validate the "happy path", the test should be something like this:
```
echo '{"workspacePath":"/home/bruna/ritchie-formulas-local", "formula": "name-of-your-formula-already-created"}' | rit build formula --stdin
```

Please forgive me for this large text. It was hard to me explain on less words.
And again, if I understood wrongly the behavior of this feature, I'll fix that, no problem :grimacing: 

Thank you for your time and your patience.

brunasilvazup:
> Hi @brunasilvazup ...
> 
> First of all, thank you for your review 
> 
> I was looking at the code, and correct me if I'm wrong, but I understood the rule this way:
> 
> When you use the command `rit create formula`, you'll have a dir with the formula's name, and inside of this dir your formula will exist.
> The goal of the build formula is take the created formula inside of its dir and just build it;
> 
> With that in mind, when you use the prompt option to build a formula, you pass through these steps:
> 
> 1. select a formula workspace;
> 2. select a formula or group:
>    2.1 select a formula name to Ritchie build it;
>    2.2 choose a group, to go down on the file tree, and then select a formula name;
> 
> When you select a formula, Ritchie will enter on chosen dir and then build the formula inside using one of those implementations of Build interface:
> 
> * local.go
> * shell.go
> * docker.go
> * make.go
> * bat.go
> 
> You can see this behavior in `build_formula.go` on line 181 when the program call `if err := b.formula.Build(info); err != nil {`.
> 
> When I developed this feature, I tryed to focus on this.
> So, `Workspace` should be the place of your formula's dir, and `Formula` should be the name of your formula, therefore the name of the formula's dir. Keeping the same behavior already implemented on prompt option.
> 
> As far I understood of the behavior of Ritchie, your test is validating a uncreated formula, so would be right you get this error: `Error: formula does not exists`
> To validate the "happy path", the test should be something like this:
> 
> ```
> echo '{"workspacePath":"/home/bruna/ritchie-formulas-local", "formula": "name-of-your-formula-already-created"}' | rit build formula --stdin
> ```
> 
> Please forgive me for this large text. It was hard to me explain on less words.
> And again, if I understood wrongly the behavior of this feature, I'll fix that, no problem 
> 
> Thank you for your time and your patience.

Hi @aronrichter,
Thanks for the explanation and you are correct in your view.
One thing that is incorrect in the development of logic is that the formula is constructed in parts using the directory structure.

> image
> In this example I have a formula `rit test build-stdin`, so I have the `test` directory and inside the` build-stdin` directory.

As you developed it only finds a formula if the directory has its full name, so it only works explicitly for formulas with a single command (rit test).
This logic is in line 263 of the `buildCommandValidator` function.

Were you able to understand the problem?
A test I suggest you do is to create a formula with 2 or 3 commands and build it via stdin.

brunasilvazup:
Hi @aronrichter, good morning, thanks for the contribution you did a great job and developed a great solution.

Unfortunately we have a new PR (https://github.com/ZupIT/ritchie-cli/pull/692) from the team that will disable the `rit build` command, which is the main change in your PR.
Thus, we will close your PR and the respective issue.

Thank you again for your dedication and if you are interested in contributing to other issues, feel free.

Issue #650, opened 2020-11-03
ERROR: unsatisfiable constraints when build a dockerfile.


afonsoaugusto:
---
name: docker build returns ERROR: unsatisfiable constraints
about: Docker build with error
labels: :beetle: bug
assignees: ''

---

**What happened**:
Hi, when execute the command `docker build -t ritchie-cli .` in root directory I receive the error: "ERROR: unsatisfiable constraints"

**What you expected to happen**:
Generate a docker image for build and run application

**How to reproduce it (as minimally and precisely as possible)**:
In root directory the project run the command `docker build .`

**Anything else we need to know?**:
My objective is using the Dockerfile for build and run the application. I using the VScode, and is my plan use the .devcontainer feature for contribute the project.

**Environment**:
- Ritchie version (use `rit --version`):
  - branch: `master`
  - commit: `8c81c544`
- Operating System:
  - Linux 5.4.0-52-generic #57-Ubuntu SMP Thu Oct 15 10:57:00 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
- Others:
  - Docker version 19.03.13, build 4484c46d9d


```bash
 ᐅ docker build -t ritchie-cli .                                                                                                                                                                                                                         
Sending build context to Docker daemon   98.5MB
Step 1/4 : ARG BUILD_IMAGE=golang:1.13-alpine
Step 2/4 : FROM $BUILD_IMAGE
 ---> 5863598a981a
Step 3/4 : RUN echo "7224aa97-1f94-4679-917d-9c7bb10074ab" > /etc/machine-id
 ---> Using cache
 ---> e6a3524349b6
Step 4/4 : RUN apk update && apk add     alpine-sdk     gettext     curl     python     py-pip     jq     git     && pip install --no-cache-dir awscli==1.16.310     && apk del py-pip     && rm -rf /var/cache/apk/* /root/.cache/pip/* /usr/lib/python2.7/site-packages/awscli/examples
 ---> Running in 299a3e320b1f
fetch http://dl-cdn.alpinelinux.org/alpine/v3.12/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.12/community/x86_64/APKINDEX.tar.gz
v3.12.1-15-gcb84081c23 [http://dl-cdn.alpinelinux.org/alpine/v3.12/main]
[...]

PR #651, opened 2020-11-03
Change apk add from python to python3.

afonsoaugusto:
Signed-off-by: Afonso Rodrigues <afonso.rodrigues@maxmilhas.com.br>

### Description

Because https://github.com/ZupIT/ritchie-cli/issues/650

### How to verify it
Execute the command `docker build .`

### Changelog

Change the package installed in Dockerfile from python to python3

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ RUN apk update && apk add \
     alpine-sdk \
     gettext \
     curl \
-    python \
+    python3 \
     py-pip \
     jq \
     git \


lucasdittrichzup (approved):

henriquemoraeszup (approved):

Dockerfile
```diff
@@ -6,7 +6,7 @@ RUN apk update && apk add \
     alpine-sdk \
     gettext \
     curl \
-    python \
+    python3 \
     py-pip \
```
brunasilvazup, line 10: With this change, we do not need change to pip3 too?

afonsoaugusto: Hi @brunasilvazup, I changed only package python because the modification is for the solution the error.
And I not change the py-pip for py3-pip because the I don't has the domain the application for the identify the use.

Maybe open the new issue to change this package.

PR #654, opened 2020-11-04
feat: add default text input to formula

### LINKED ISSUE #565: Add default field visualization to survey inputs
When an input has a default value it must be shown at the end of the question in parentheses.

The input library (survey) already has this attribute, the necessary change is to change the interfaces of the inputs by adding the parameter as optional and adding the information in the desired places.

**Why is this needed**:
To make the default value attribute more visible to the user.
Improving usability.
Currently, URL inputs already have this logic.

image

lucasdittrichzup:
## Description

Add a **type of text input to formulas** that shows (and uses) the **default value** (if the input has one) in parentheses at the end of the question

default

This was necessary to make it more evident to the user that, by not informing value in an input that has a default value, it will be used.

Resolve #565

##  How to verify it

Run a formula that has input text with default value.

##  Changelog

Add a default text input to formulas.


GuillaumeFalourd:
Hi @lucasdittrichzup , looking at the gif, it looks that the `default` word on the command line is part of the formula command line.
Wouldn't it be better to inform it through a flag, such as `--default` for more clarity ?

victorschumacherzup:
> Hi @lucasdittrichzup , looking at the gif, it looks that the `default` word on the command line is part of the formula command line.
> Wouldn't it be better to inform it through a flag, such as `--default` for more clarity ?

Hi @GuillaumeFalourd , this pull request is about #565 , the flag `--default` will be implemented in another one.

lucasdittrichzup:
> Hi @lucasdittrichzup , looking at the gif, it looks that the `default` word on the command line is part of the formula command line.
> Wouldn't it be better to inform it through a flag, such as `--default` for more clarity ?

Hi @GuillaumeFalourd, there are cases where we use the **default value** of the formula **config.json field**, but we do not make this clear to the user. 

Thus, we decided to present the default value through the **default option of the survey library**. As @victorschumacherzup commented, the **default flag** is another implementation.

pkg/formula/input/prompt/text_default.go
```diff
@@ -0,0 +1,55 @@
...
+		},
+	}
+
+	if len(i.Tutorial) > 0 {
+		input.Help = i.Tutorial
+	}
+
+	if len(i.Default) > 0 {
+		input.Default = i.Default
+	}
```
henriquemoraeszup, line 48: Are the checks required? If they are an empty string it will only be set to an empty string anyway, right?

henriquemoraeszup (approved):
Good job, very concise 😄 

Just one comment above

pkg/formula/input/prompt/text_default.go
lucasdittrichzup: in fact the survey is even more intelligent, because if an empty string is passed to the help or the default, it doesn't even show the values

victorschumacherzup:
tested the functionalitty and it works. congrats

brunasilvazup (approved):

PR #666, opened 2020-11-11
Feature/split_local

kaduartur:
##### Local repos
Now when we create a formula it will be built and sent to a specific repository according to the name of your workspace with the prefix **"local"**, for example:
if we have a workspace with the name my-work-name.

```go
Workspace {
  Name: "my-work-name",
  Dir: "/home/user/my-workspace-path"
}
```
we will have a repository called "local-my-work-name" that will have the highest priority.

Remembering that with this modification we gained some commands for local repositories such as `rit list repo`, `rit set repo-priority`, and `rit delete repo`

---
##### Repo priority
A problem with the repository priority has been fixed, now when setting a repository with negative priority (-1) this repo will have the highest priority having its priority set to 0. When setting a repo with a priority higher than the amount of repository you have added that will have the lowest priority. for example, if we have a list with 5 repositories and you set the priority of some to 99 this will have priority 4 where this is the lowest priority among the repositories.

##### Local build modifications

Today, by default, if a formula doesn't have the /bin folder, it means it hasn't been built yet, so Ritchie builds it automatically when the user types the formula command. The logic for executing build scripts with Makefile, build.sh, or build.bat has been removed from the local build to keep in one place, so we won the build with docker for local formulas.


pkg/cmd/set_priority.go
```diff
@@ -97,12 +97,19 @@ func (s setPriorityCmd) runFunc() CommandRunnerFunc {
...
-		err = s.SetPriority(repo.Name, int(priority))
-		if err != nil {
+		if err := s.SetPriority(repo.Name, int(priority)); err != nil {
 			return err
 		}
 
 		successMsg := fmt.Sprintf(newRepositoryPriority, repoName, priority)
+		if int(priority) > repositories.Len() {
+			successMsg = fmt.Sprintf("Now %q repository has the least priority", repoName)
+		}
```
brunasilvazup, line 107: Is this information always true?

In my view, only comparing all the priorities of the repositories to know the last one, because the priority value is free right?

kaduartur: it's not like that anymore, now the maximum priority is the size of the repository array

pkg/formula/repo/list_write_creator_test.go
```diff
@@ -43,30 +43,35 @@ func TestNewListCreator(t *testing.T) {
...
 				RepositoryCreator: repoCreator,
+				RepositoryWriter:  repoWrite,
 			},
 		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			if got := NewListCreator(tt.in.repoList, tt.in.repoCreate); !reflect.DeepEqual(got, tt.want) {
+			if got := NewListWriteCreator(tt.in.repoList, tt.in.repoCreate, tt.in.repoWrite); !reflect.DeepEqual(got, tt.want) {
 				t.Errorf("NewListCreator() = %v, want %v", got, tt.want)
```
brunasilvazup, line 75: Typo "NewListCreator" -> "NewListWriteCreator"

brunasilvazup: And the name of test too (I couldn't put the comment there #humiliated)

kaduartur: **#humiliated** LOL

pkg/cmd/set_priority.go
brunasilvazup: Yes, I understood.
But i think this can this cause confusion for him, if he sets priority 99 for repoA and then checks priorities will appear 3 (if he has 4 repos for example) and not what he set.

PS: I think it's a great change, but I think that could be explicit for the user.

pkg/formula/repo/priority_setter.go
```diff
@@ -17,64 +17,79 @@
...
+	}
 
-	bytes, err := json.MarshalIndent(repos, "", "\t")
-	if err != nil {
-		return err
+	for i = index; i < priority; i++ { // Move repos to front
+		r := repos[i+1]
+		r.Priority = i
+		repos[i] = r
 	}
```
brunasilvazup, line 77: What do you think about putting these two for's in separate functions, I had a hard time understanding what they were doing.
I think putting a function with the right name already helps.

pkg/formula/repo/priority_setter.go
```diff
@@ -17,64 +17,79 @@
...
-			repos[i].Priority++
+			repo = repos[i]
+			index = i
+			break
 		}
 	}
 
-	sort.Sort(repos)
+	var i int
+	for i = index; i > priority; i-- { // Move repos to back
```
brunasilvazup, line 67: Back and front could be confused, what do you think about `head` and `tail`

pkg/formula/repo/priority_setter.go
kaduartur: Do you think it makes sense to create two more copies of the repositories slice?

pkg/cmd/set_priority.go
kaduartur: I will improve this message by explaining why such a priority

pkg/formula/repo/priority_setter.go
```diff
@@ -17,64 +17,79 @@
...
+		repos[i] = r
+	}
 
-	bytes, err := json.MarshalIndent(repos, "", "\t")
-	if err != nil {
-		return err
+	for i = index; i < priority; i++ { // Move repos to front
+		r := repos[i+1]
+		r.Priority = i
+		repos[i] = r
```
henriquemoraeszup: Why not insert it at the priority and iterate the whole array overwriting the priorities? Seems less complex

pkg/formula/repo/priority_setter_test.go
```diff
@@ -65,24 +64,175 @@ func TestSetPriorityManager_SetPriority(t *testing.T) {
...
+				}(),
+			},
+			args: args{
+				repoName: "repo-2-local",
+				priority: -1,
+			},
+			wantErr: false,
+		},
+		{
+			name: "Setting negative repo priority",
```
henriquemoraeszup, line 142: this test has the same name as the previous one

pkg/formula/repo/priority_setter.go
henriquemoraeszup: I would make an insertion and iterate the whole array resetting the priorities

henriquemoraeszup: You can actually make this in a single loop and eliminate these two + the one before

henriquemoraeszup:
Is the base branch you are pointing to intentional?

pkg/formula/repo/priority_setter.go
kaduartur: If I do that I may end up overwriting some repository.

pkg/formula/builder/local.go
```diff
@@ -17,40 +17,41 @@
...
-	file stream.FileWriteReadExister,
-	tree formula.TreeGenerator,
+	repo formula.RepositoryAdder,
 ) LocalManager {
-	return LocalManager{ritHome: ritHome, dir: dir, file: file, tree: tree}
+	return LocalManager{ritHome: ritHome, dir: dir, repo: repo}
 }
 
 func (m LocalManager) Build(info formula.BuildInfo) error {
+	repoName := fmt.Sprintf("%s-local", info.Workspace.Name)
```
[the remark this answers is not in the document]
kaduartur: Good idea, I'll modify this

pkg/formula/repo/list_write_creator_test.go
victorschumacherzup: LOL

pkg/formula/repo/priority_setter.go
kaduartur: This was discussed at a meeting and was agreed with this logic

pkg/formula/repo/priority_setter.go
kaduartur: This was discussed at a meeting and was agreed with this logic

pkg/formula/builder/local_test.go
```diff
@@ -56,101 +63,65 @@ func TestBuild(t *testing.T) {
...
 			want: nil,
 		},
 		{
 			name: "create dir error",
 			in: in{
 				formulaPath: formulaPath,
-				fileManager: fileManager,
 				dirManager:  dirManagerMock{createErr: errors.New("error to create dir")},
-				tree:        defaultTreeManager,
+				repo:        repoAdder,
```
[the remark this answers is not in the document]
kaduartur: Eagle eyes 🦅 👀

victorschumacherzup (approved):

PR #667, opened 2020-11-12
removed grpc files

victorschumacherzup:
### Description
 Removing grpc files since we don't use it


henriquemoraeszup (approved):
O love deleting code

brunasilvazup (approved):

PR #677, opened 2020-11-16
Adding --force-build flag

### LINKED ISSUE #668: Add --force-build flag to run a formula
## Description

Create a flag of `--force-build` to clear both the cache and the `bin` of the formula to force a rebuild

gabriel-pinheiro:
Signed-off-by: Gabriel Pinheiro <gabrielctpinheiro@gmail.com>

### Description
You can now run formulas with a `--force-build` (or `-b`) flag that will force it to be built (ok haha) regardless of changes without prompting the user with an extra question

### How to verify it
Run any formula that exists locally in a repository with --force-build or -b and check the build spinner before the execution

### Changelog
Added `--force-build` (or `-b`) flag

force-build

Closes #668


pkg/formula/runner/pre_run_builder.go
```diff
@@ -22,37 +22,60 @@ import (
...
+	workspace, err := b.anyWorkspace(relativePath)
+	if err != nil {
+		fmt.Println(prompt.Yellow(messageListError))
+		return
+	}
+	if workspace == nil {
+		fmt.Println(prompt.Yellow(messageNoLocalSource))
+		return
+	}
+
```
henriquemoraeszup, line 72: I think we need to remove the `bin` folder and the input cache for the build to work properly, can you confirm @kaduartur ?

gabriel-pinheiro: It looks like the `bin` folder is already being removed in the line below that on b.buildOnWorkspace -> b.builder.Build -> (now on `formula/builder/local.go`, `Build` func) m.buildFormulaBin -> m.dir.Remove (local.go L78)

I'll take a look at how that cache is stored and push a commit with that update, thanks for pointing that out!

pkg/cmd/formula.go
```diff
@@ -213,7 +226,7 @@ func (f FormulaCommand) execFormulaFunc(repo, path string) func(cmd *cobra.Comma
 			Flags:   flags,
 		}
 
-		if err := f.formula.Execute(exe); err != nil {
+		if err := f.formula.Execute(exe, forceBuild); err != nil {
```
kaduartur, line 229: could you create a new field in `formula.ExecuteData` or use the flags field itself to capture the value of` --force-build`?
So you don't break the interface contract.

pkg/formula/runner/pre_run_builder.go
kaduartur: I think this should only work for formulas that are in a repository. Because many of the formulas don't have any workspace.

- remove the `/bin` folder
- remove cache files

The Ritchie by default will build the formula because there is no longer a `/bin` folder

kaduartur (changes requested):

pkg/formula/runner/pre_run_builder.go
gabriel-pinheiro: Ohh got it! I'll push these changes.

Also, when using the `--force-build`, should Ritchie only rebuild the formula already on `~/.rit/repos` or also override the sources in there from the workspace (if it exists in one)?

GuillaumeFalourd: I wouldn't override the sources files from `formulas workspaces`. Only the files in the `~/.rit/repos` 🤔 

What scenario do you see where it could be interesting to override the sources files ? (I may be missing something)

gabriel-pinheiro: Ohh I mean to override the `~/.rit/repos` with the workspace source (instead of only building what's already on `~/.rit/repos`)

GuillaumeFalourd: Sounds like a good idea to me @gabriel-pinheiro 👍🏼

gabriel-pinheiro:
I just pushed the changes to build the formula even when it doesn't exist on any workspace (by just deleting the bin dir). I kept the previous code that would update the rit home with the workspace source.

I also pushed the refactor to get the force-build flag from ExecuteData.

I just need to add some more tests (for the new logic) and clear the formula cache in the --force-build and the PR will be ready to approve :)

henriquemoraeszup:
Wassup @gabriel-pinheiro, is this PR still a draft? I am the release manager this week, if you require any help or have any questions let me know =]

gabriel-pinheiro:
Hey henrique, I recently got some help from Bruna and Kadu, I'll be able to push the changes now 😄

lucasdittrichzup:
Hi @gabriel-pinheiro, is this PR still a draft?

brunasilvazup:
Hi @gabriel-pinheiro, as some time has passed since the beginning of this PR we decided to close it for now.
If you want to continue or stop this PR feel free.

Thank you for your contribution and if you want to talk we are available.

PR #685, opened 2020-11-19
Use latest version of repository in rit add repo stdin

brunasilvazup:
### Description

In this PR, we changed the behavior of rit add repo when using stdin and not including the version.

Now ritchie will default to the latest version of the repository.

image

### How to verify it

Clone this PR, update rit with this version and run the following command:

`echo '{"provider":"Github", "name":"repoName", "url":"https://github.com/ZupIT/ritchie-formulas", "token": null, "priority":1}' | rit add repo --stdin`

### Changelog

Use latest version of repository in rit add repo stdin


pkg/cmd/list_repo.go
```diff
@@ -81,7 +82,10 @@ func (lr listRepoCmd) printRepos(repos formula.Repos) {
 	for _, repo := range repos {
 		latestTag := "0.0.0"
 		if !repo.IsLocal {
-			latestTag = lr.getLatestTag(repo)
+			latestTag = lr.detail.LatestTag(repo)
+			if len(latestTag) == 0 {
+				latestTag = "Couldn't get that information"
```
henriquemoraeszup, line 87: where is this variable used?

pkg/formula/repo/detail_test.go
```diff
@@ -0,0 +1,101 @@
...
+					Priority: 0,
+				},
+				funcLatestTag: func(info git.RepoInfo) (git.Tag, error) {
+					return git.Tag{Name: "1.0.0"}, nil
+				},
+			},
+			want: "1.0.0",
+		},
+		{
+			name: "Return version nill when get latest returns error",
```
henriquemoraeszup, line 69: nitpick
```suggestion
			name: "Return version nil when get latest returns error",
```

pkg/formula/repo/detail_test.go
```diff
@@ -0,0 +1,101 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			defaultGitRepositoryMock.latestTag = tt.fields.funcLatestTag
+			repoProviders.Add("Github", formula.Git{Repos: defaultGitRepositoryMock, NewRepoInfo: github.NewRepoInfo})
+			dm := NewDetail(tt.fields.repoProviders)
+
+			tag := dm.LatestTag(tt.fields.repo)
+
+			if tag != tt.want {
+				t.Errorf("TestNewDetailLatestTag() receive = %v, expected %v", tag, tt.want)
+			}
```
henriquemoraeszup, line 98: werent we using testify?

pkg/cmd/add_repo.go
```diff
@@ -199,14 +201,18 @@ func (ad addRepoCmd) runPrompt() CommandRunnerFunc {
...
 		r := formula.Repo{}
 
-		err := stdin.ReadJson(os.Stdin, &r)
+		err := stdin.ReadJson(cmd.InOrStdin(), &r)
 		if err != nil {
 			return err
 		}
 
+		if r.Version.String() == "" {
+			latestTag := ad.detail.LatestTag(r)
```
henriquemoraeszup, line 212: is this method making a request? I was wondering if we could solve this with a single request and get the last tag issued from the list as latest, does that make sense?

pkg/cmd/list_repo.go
brunasilvazup: In line 91.

pkg/formula/repo/detail_test.go
brunasilvazup: Yes,
I was in doubt if I changed the asserts or did a pr changing the entire file, but I think I can already use the testify assert here.

pkg/cmd/add_repo.go
brunasilvazup: Yes, I will evaluate, but there would have to be different implementations for this need.

brunasilvazup: Hi @henriquemoraeszup ,
I searched and did not find a method of doing these two operations as a single request.

I think that using two requests is not a problem due to logic.
These two requests will only be made when the entry is via stdin and the user does not inform the version.
So it is natural that we have some work due to this lack of information.

pkg/cmd/add_repo_test.go
```diff
@@ -179,10 +226,22 @@ func Test_addRepoCmd_runPrompt(t *testing.T) {
...
 			)
-			o.PersistentFlags().Bool("stdin", false, "input by stdin")
-			if err := o.Execute(); (err != nil) != tt.wantErr {
-				t.Errorf("init_runPrompt() error = %v, wantErr %v", err, tt.wantErr)
+			if tt.fields.stdin == "" {
+				cmd.PersistentFlags().Bool("stdin", false, "input by stdin")
+
+				if err := cmd.Execute(); (err != nil) != tt.wantErr {
+					t.Errorf("NewAddRepoCmd runPrompt error = %v, wantErr %v", err, tt.wantErr)
+				}
```
henriquemoraeszup, line 236: you can take this snippet out of the `if` clause since it is like the one below and assert `noError` or assert `empty` depending if you are expecting an error or not. We should already know from the name of the test whether it is a stdin or prompt test

pkg/cmd/list_repo_test.go
```diff
@@ -55,6 +56,11 @@ func TestListRepoRunFunc(t *testing.T) {
 				},
 				Tutorial: TutorialFinderMockReturnDisabled{},
 				Repos:    defaultGitRepositoryMock,
+				Detail: DetailManagerCustomMock{
+					latestTag: func(repo formula.Repo) string {
+						return "2.0.0"
+					},
+				},
```
henriquemoraeszup, line 63: How about declaring two different mocks and passing them? One for the empty tag one for the 2.0.0

pkg/cmd/list_repo_test.go
```diff
@@ -187,3 +213,11 @@ type RepositoryListerCustomMock struct {
 func (m RepositoryListerCustomMock) List() (formula.Repos, error) {
 	return m.list()
 }
+
+type DetailManagerCustomMock struct {
+	latestTag func(repo formula.Repo) string
+}
+
+func (d DetailManagerCustomMock) LatestTag(repo formula.Repo) string {
+	return d.latestTag(repo)
```
henriquemoraeszup, line 222: Should we use testify mocks?

pkg/prompt/text_validator.go
```diff
@@ -40,6 +42,7 @@ func (SurveyTextValidator) Text(name string, validate func(interface{}) error, h
 	} else {
 		validationQs[0].Prompt = &survey.Input{Message: name}
 	}
+	fmt.Println("text_validador: ", 1)
```
henriquemoraeszup, line 45: is this intentional?

pkg/cmd/add_repo_test.go
brunasilvazup: The need for this if is due to having unique tests for stdin and when they are executed lines 238-240 are needed.

An alternative would be to create two different cmds and add an if to see what will be executed.

Do you think this is a better solution or should you separate it into two different test functions?

pkg/prompt/text_validator.go
brunasilvazup: Oh no, thanks.

pkg/cmd/list_repo_test.go
brunasilvazup: This is resolved with testify mocks, do you think it looks good?

pkg/cmd/add_repo_test.go
brunasilvazup: I had got it wrong, it can certainly be simplified.
I improved with a default testing solution, what do you think?

henriquemoraeszup (approved):
Niice, tests are looking good 🚀

lucasdittrichzup (approved):

brunasilvazup:
\merge qa

Issue #554, opened 2020-09-25
Switch the command context for environment


leticiasilvazup:
**What would you like to be added**:

Work on the context command switching it to environments,

**Why is this needed**:

To match the market and Ritchie Cloud

### Tasks

- [x] Remove the current `env` package (#681)
- [x] Move the `rcontext` to new `env` package (#687)
- [x] Edit all variable and methods named `context` inside `rcontext` package (#687)
- [x] Create command `rit set env` (#688)
- [x] Deprecate command `rit set context` (#688)
- [x] Create command `rit show env` (#691)
- [x] Deprecate command `rit show context` (#691)
- [x] Create command `rit delete env` (#694)
- [x] Deprecate command `rit delete context` (#694)
- [x] Convert the file `contexts` to `envs` (#697)
- [ ] Update documentation

kaduartur:
### Tasks

- [ ] Remove the current `env` package
- [ ] Move the `rcontext` to new `env` package
- [ ] Edit all variable and methods named `context` inside `rcontext` package
- [ ] Create command `rit set env`
- [ ] Create command `rit delete env`
- [ ] Create command `rit show env`
- [ ] Deprecate command `rit set context`
- [ ] Deprecate command `rit delete context`
- [ ] Deprecate command `rit show context`
- [ ] Convert the file `contexts` to `envs`

PR #688, opened 2020-11-23
Create new command rit set env

kaduartur:
### Description
The `rit set context` command has been deprecated and we will now use the new` rit set env` command which has the same functionality as the old command.

### How to verify it
It's a small part of the #554 issue, a separate pull request was created to facilitate code review


pkg/cmd/set.go
```diff
@@ -17,17 +17,31 @@
...
+		Example:   "rit set env",
+		ValidArgs: []string{"env", "credential", "formula-runner", "repo-priority"},
 		Args:      cobra.OnlyValidArgs,
 	}
+
+	deprecatedMsg := fmt.Sprintf(
+		`you can now use the "%v" command for the same purpose as the "%v" command.`,
+		prompt.Bold("rit set env"),
+		prompt.Bold("rit set context"),
+	)
```
victorschumacherzup, line 42: maybe `%q` can be useful here

victorschumacherzup (approved):

pkg/cmd/set.go
kaduartur: the format **"%q"** does not work because of the `prompt.Bold` inserts bold text for this text and prints invalid characters

PR #689, opened 2020-11-23
Command based text to help.json

victorschumacherzup:
### Description

Small change adding a command based text to `help.json`

### How to verify it

Use `rit create formula` and check the `help.json` files.

### Changelog
added a command based text to help.json


brunasilvazup (approved):

pkg/formula/creator/creator.go
```diff
@@ -151,17 +151,22 @@ func (c CreateManager) applyLangTemplate(lang, formulaPath, workspacePath string
...
+	var commands string
+	for i := 0; i < dirsLen; i++ {
 		d := dirs[0 : i+1]
 		tPath := filepath.Join(workSpacePath, filepath.Join(d...))
 		helpPath := filepath.Join(tPath, template.HelpFileName)
 		if !c.file.Exists(helpPath) {
 			folderName := filepath.Base(tPath)
-			tpl := strings.ReplaceAll(template.HelpJson, "{{folderName}}", folderName)
-			help := formula.Help{}
+			commands += folderName + " "
```
kaduartur, line 162: Why add a space here?

pkg/formula/creator/template/default.go
```diff
@@ -19,8 +19,13 @@ package template
 const (
 	HelpJson = `
 {
-    "short": "short help placeholder for {{folderName}}",
-    "long": "long help placeholder for {{folderName}} used by index page and -h"
+    "short": "{{command}}commands",
```
kaduartur, line 22: could you use the Golang `fmt`?

kaduartur (changes requested):

pkg/formula/creator/creator.go
victorschumacherzup: it goes between the commands, without it the command is built like awscreate

pkg/formula/creator/template/default.go
victorschumacherzup: sure!

pkg/formula/creator/template/default.go
```diff
@@ -19,8 +19,13 @@ package template
...
+    "long": "{{command}}commands"
+}`
+	FormulaHelpJson = `
+{
+    "short": "{{command}}formula",
```
henriquemoraeszup, line 27: If you already add a space on the template you will be able to  join the commands more elegantly
```suggestion
    "short": "{{command}} formula",
```

pkg/formula/creator/creator.go
henriquemoraeszup, line 162: Instead of concatenating with empty space, why not join the `dirs` array?
```
fieldPath := filepath.Join(fields[0:i]...)
commands := strings.Join(fields[0:i], " ")
```

pkg/formula/creator/creator.go
```diff
@@ -151,17 +151,22 @@ func (c CreateManager) applyLangTemplate(lang, formulaPath, workspacePath string
...
 func (c CreateManager) createHelpFiles(formulaCmdName, workSpacePath string) error {
 	dirs := strings.Split(formulaCmdName, " ")
-	for i := 0; i < len(dirs); i++ {
+	dirsLen := len(dirs)
+	var commands string
+	for i := 0; i < dirsLen; i++ {
 		d := dirs[0 : i+1]
 		tPath := filepath.Join(workSpacePath, filepath.Join(d...))
 		helpPath := filepath.Join(tPath, template.HelpFileName)
 		if !c.file.Exists(helpPath) {
```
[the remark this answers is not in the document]
victorschumacherzup: thanks for recommending this reading
done 11bbf5d7fa3b9e9521978d207408fc6d8e35d3c7

kaduartur (approved):
LGTM 🥇

pkg/formula/creator/template/default.go
victorschumacherzup: yes! done in a68d9d602086fa25241f8b7ee7ddd9fe28d4e853

pkg/formula/creator/creator.go
victorschumacherzup: changed in a68d9d602086fa25241f8b7ee7ddd9fe28d4e853

henriquemoraeszup (changes requested):
I think we are missing the tests to check for these files

brunasilvazup (dismissed):

pkg/formula/creator/creator_test.go
```diff
@@ -204,9 +207,20 @@ func TestCreator(t *testing.T) {
...
 			}
+
+			if in.helpPath != "" {
+				bytes, _ := fileManager.Read(in.helpPath)
+				help := formula.Help{}
+				_ = json.Unmarshal(bytes, &help)
+				if help.Short == "" || help.Long == "" {
+					t.Error("help.json should not be empty")
+				}
+			}
```
henriquemoraeszup, line 223: Can you just check for the exact expected test? Also you can apply testify for your assertions

pkg/formula/creator/creator.go
```diff
@@ -151,28 +152,33 @@ func (c CreateManager) applyLangTemplate(lang, formulaPath, workspacePath string
...
+		commands := strings.Join(d, " ")
+		complement := "commands"
+		if i == dirsLen-1 {
+			complement = "formula"
+		}
+		tpl := fmt.Sprintf(template.HelpJson, commands, complement, commands, complement)
+		help := formula.Help{}
+		if err := json.Unmarshal([]byte(tpl), &help); err != nil {
+			return err
+		}
```
henriquemoraeszup, line 173: you can avoid this unmarshall step by just setting the help properties, like
`help.short = fmt.Sprintf("%s %s", commands, complement)`

pkg/formula/creator/creator_test.go
```diff
@@ -204,9 +209,21 @@ func TestCreator(t *testing.T) {
 			creator := NewCreator(treeMan, tt.in.dir, tt.in.file, tt.in.tplM)
 			out := tt.out
 			got := creator.Create(in.formCreate)
-			if (got != nil && out.err == nil) || got != nil && got.Error() != out.err.Error() || out.err != nil && got == nil {
+			if (got != nil && out.err == nil) ||
+				got != nil && got.Error() != out.err.Error() ||
+				out.err != nil && got == nil {
 				t.Errorf("Create(%s) got %v, want %v", tt.name, got, out.err)
 			}
```
henriquemoraeszup, line 216: How about making testify standard? Also, this boolean logic seems complex and hard to understand
```suggestion
			if (out.err == nil) {
			      require.NoError(t, got)
			} else {
			     require.NotNil(t, out.err)
			     require.Equal(t, err.Error(), out.err.Error())
			}
```

pkg/formula/creator/creator_test.go
```diff
@@ -204,9 +209,21 @@ func TestCreator(t *testing.T) {
...
+
+			if in.helpPath != "" {
+				bytes, _ := fileManager.Read(in.helpPath)
+				help := formula.Help{}
+				_ = json.Unmarshal(bytes, &help)
+				errorMsg := "help.json should not be empty"
+				expected := "scaffold generate test_go formula"
+				assert.Equal(t, expected, help.Short, errorMsg)
```
henriquemoraeszup, line 224: Could you check if a wrong assert is actually failing the test? I think assert only returns a boolean without failing it. `require` seems to do the job better

pkg/formula/creator/template/default.go
```diff
@@ -19,8 +19,8 @@ package template
 const (
 	HelpJson = `
 {
-    "short": "short help placeholder for {{folderName}}",
-    "long": "long help placeholder for {{folderName}} used by index page and -h"
+    "short": "%s %s",
+    "long": "%s %s"
 }`
```
henriquemoraeszup, line 24: If you set the properties as suggested above, you will not need this template

pkg/formula/creator/creator_test.go
victorschumacherzup: The test fails with an `assert` as well. I really don't know the difference between them

pkg/formula/creator/template/default.go
victorschumacherzup: yes! thanks

pkg/formula/creator/creator_test.go
victorschumacherzup: sure! removed

pkg/formula/creator/creator.go
victorschumacherzup: nice idea! :bulb: 
 thanks

pkg/formula/creator/creator_test.go
```diff
@@ -204,8 +209,21 @@ func TestCreator(t *testing.T) {
...
+			}
+
+			if in.helpPath != "" {
+				bytes, _ := fileManager.Read(in.helpPath)
+				help := formula.Help{}
+				_ = json.Unmarshal(bytes, &help)
+				errorMsg := "help.json should not be empty"
+				expected := "scaffold generate test_go formula"
+				require.Equal(t, expected, help.Short, errorMsg)
+				require.Equal(t, expected, help.Short, errorMsg)
```
henriquemoraeszup, line 226: I think here you wanted to compare with `help.Long`

henriquemoraeszup (dismissed):
Nice! Just one comment to fix, but I am good to go 🚀

pkg/formula/creator/creator_test.go
victorschumacherzup: yes, fixed it on e940471

henriquemoraeszup (approved):

brunasilvazup (approved):
:+1:

PR #704, opened 2020-12-01
Add step to verify repo existence

brunasilvazup:
### Description

This PR add a verification in `rit add repo` (stdin and prompt) to verify if the formula repo has already been added.

* This verification happens after receiving the url and the version.

> Prompt
image

>Stdin
image

>Other version, don't presents error
image

### How to verify it
Clone this PR and add a repo existent, (commons repo for example).

### Changelog

Add step to verify repo existence


pkg/cmd/add_repo_test.go
```diff
@@ -34,12 +34,30 @@ func TestAddRepoCmd(t *testing.T) {
...
+		Priority: 2,
+	}
+	repoListerPopulated := repoListerAdderCustomMock{
+		add: func(d formula.Repo) error {
+			return nil
+		},
+		list: func() (formula.Repos, error) {
+			return formula.Repos{*repoTest}, nil
+		},
+	}
```
henriquemoraeszup, line 52: can we use testify mocks? Its ok that we duplicate a mock for the same purpose, we can remove instances of `repoListerAdderCustomMock` in the future

lucasdittrichzup (approved):

pkg/cmd/add_repo_test.go
brunasilvazup: Yes, i will change this.

henriquemoraeszup (dismissed):
Nice test refact ❤️

henriquemoraeszup:
Hint, maybe to hit the coverage you need a case in which `exitsRepo` returns false

brunasilvazup:
> exitsRepo

Yes, i did in the last test `" Return error when user add a repo existent "`

pkg/cmd/add_repo_test.go
```diff
@@ -54,157 +103,217 @@ func TestAddRepoCmd(t *testing.T) {
...
-					list: func(name string, items []string) (string, error) {
-						return "Github", nil
-					},
+				repo:               returnOfRepoListerAdder{errAdd: nil, reposList: formula.Repos{}, errList: nil},
+				InputTextValidator: returnWithStringErr{"mocked text", nil},
+				InputPassword:      returnWithStringErr{"s3cr3t", nil},
+				InputURL:           returnWithStringErr{"http://localhost/mocked", nil},
+				InputBool:          returnOfInputBool{true, nil},
+				InputList:          []returnOfInputList{{response: "Github", err: nil}},
+				stdin:              "{\"provider\": \"github\", \"name\": \"repo-name\", \"version\": \"\", \"url\": \"https://url.com/repo\", \"token,omitempty\": \"\", \"priority\": 5, \"isLocal\": false}\n",
```
victorschumacherzup, line 298: What you think about put the string between `? So the slashes can be removed

victorschumacherzup: Since it is repeated on 284, a const will be nice

brunasilvazup: Hi, it's a good idea, i will be change that, tks (:

brunasilvazup: But this entry hasn't version and in entry of line 284 has, so i will be change only ` ok?

victor-schumacher (dismissed):

victorschumacherzup (approved):

pkg/cmd/add_repo_test.go
```diff
@@ -54,157 +103,217 @@ func TestAddRepoCmd(t *testing.T) {
...
+		{
+			name: "Run with success when user add a new commons",
+			fields: fields{
+				repo:               returnOfRepoListerAdder{errAdd: nil, reposList: formula.Repos{formula.Repo{Provider: "Github", Name: formula.RepoCommonsName}}, errList: nil},
+				InputTextValidator: returnWithStringErr{formula.RepoCommonsName.String(), nil},
+				InputPassword:      returnWithStringErr{"s3cr3t", nil},
+				InputURL:           returnWithStringErr{"http://localhost/mocked", nil},
+				InputBool:          returnOfInputBool{true, nil},
+				InputList:          []returnOfInputList{{response: "Github", err: nil}},
+				tutorialStatus:     "disabled",
```
henriquemoraeszup, line 125: Hey! O love this approach, I was just thinking, could we somehow have a default return in place and only override the returns we want to change? I feel this way we would end up with less repeated code and on each test, we would explicitly have only the mocked field with a different side effect. What do you think?

internal/mocks/mocks.go
```diff
@@ -42,3 +46,117 @@ func (d *DetailManagerMock) LatestTag(repo formula.Repo) string {
...
+
+	return args.Error(0)
+}
+
+type TutorialFindSetterMock struct {
+	mock.Mock
+}
+
+func (t *TutorialFindSetterMock) Find() (rtutorial.TutorialHolder, error) {
+	args := t.Called()
```
henriquemoraeszup, line 71: Bruna, I was thinking, we might end up with too many mocks. Also the only other folder in `internal` is `proto`, which we are not using. What do you think of a 1st-level folder `mocks` with different scoped files for them? All from `mocks` package but more like
```
mocks
    input.go (bool, list, int)
    formula.go
    actions.go (add, delete, find)
    cmd.go
```

We might end up with a problem of categorizing mocks properly, so I am not sure this is still the best approach. But I am still in favor of moving the `mocks` folder up one level 😄

pkg/cmd/add_repo_test.go
brunasilvazup: Do you say something similar to that?
-> 
https://github.com/ZupIT/ritchie-cli/blob/e09272256bd7f000a8aaeeb0eb82ee0a37d9f988/pkg/formula/input/flag/flag_test.go#L144

henriquemoraeszup: That could definitely work, its sad that golang does not have coalesce operator 😢

internal/mocks/mocks.go
henriquemoraeszup: sounds good

pkg/cmd/add_repo.go
```diff
@@ -241,6 +255,15 @@ func (ad addRepoCmd) repoNameValidator(text interface{}) error {
 	return nil
 }
 
+func exitsRepo(urlToAdd, versionToAdd string, repos formula.Repos) bool {
```
henriquemoraeszup, line 258: nitpick
```suggestion
func existsRepo(urlToAdd, versionToAdd string, repos formula.Repos) bool {
```

henriquemoraeszup (dismissed):

henriquemoraeszup (approved):

PR #705, opened 2020-12-01
Check repository structure

### LINKED ISSUE #698: Check repository structure on RIT ADD REPO command
## What would you like to be added

- I would like the rit add repo to check if the imported repository is a ritchie repository that contains formulas.

## Why is this needed

- Today, it's possible to use the `rit add repo` command to add any Github or Gitlab repository. If the repository isn't a formulas repository, why allow to add it to the `.rit/repos` directory ?

### Example 

`rit add repo` with https://github.com/GuillaumeFalourd/developers-cheat-sheets repository 
(This repo had a release tag at that time) 

example

matheussn:
### Description

I implemented a validation when adding repositories, so that the user can only add repositories that have valid formulas
closes #698

### How to verify it
To validate, you can try to add any repository that is not a Ritchie formula repository, using the `rit add repo` command. The same, should return an error to the user with a message saying that the chosen repository does not have formulas.

### Changelog

Validation in the rit add repo command, to add only formula repositories


pkg/cmd/add_repo.go
```diff
@@ -180,6 +183,9 @@ func (ad addRepoCmd) runPrompt() CommandRunnerFunc {
 		}
 
 		if err := ad.repo.Add(repository); err != nil {
+			if err := ad.repoDeleter.Delete(repository.Name); err != nil {
```
kaduartur, line 186: I think that this logic should be within the "business" layer, more specifically within the Add function.

pkg/cmd/add_repo.go
```diff
@@ -214,6 +220,9 @@ func (ad addRepoCmd) runStdin() CommandRunnerFunc {
 		}
 
 		if err := ad.repo.Add(r); err != nil {
+			if err := ad.repoDeleter.Delete(r.Name); err != nil {
```
kaduartur, line 223: I think that this logic should be within the "business" layer, more specifically within the Add function.

pkg/formula/repo/adder.go
```diff
@@ -77,6 +78,10 @@ func (ad AddManager) treeGenerate(repo formula.Repo) error {
 		return err
 	}
 
+	if len(tree.Commands) == 0 {
+		return errors.New("the selected repository has no formulas")
```
kaduartur, line 82: Here you can call the delete repository function and then return the error

kaduartur (changes requested):

pkg/formula/repo/adder.go
matheussn: Makes sense!
Could I ignore the `delete` function error?
Or do you think it is better to keep the same rule that is in the `add_repo.go` file?

brunasilvazup: At this point in the logic he has already created the repository (line 63), so it is good to delete it here.

Or you can evaluate if we could put this if of the tree first, @kaduartur think this is a good approach?

henriquemoraeszup:
Hello @matheussn just checking if everything is ok, if you need any support let me know, I'm the designated release manager for this week 😃

pkg/formula/repo/adder.go
kaduartur: > Makes sense!
> Could I ignore the `delete` function error?
> Or do you think it is better to keep the same rule that is in the `add_repo.go` file?

good practices in Golang recommend not to ignore errors. [Handle-errors](https://github.com/golang/go/wiki/CodeReviewComments#handle-errors)

kaduartur: > At this point in the logic he has already created the repository (line 63), so it is good to delete it here.
> 
> Or you can evaluate if we could put this if of the tree first, @kaduartur think this is a good approach?

when using the delete repository function we have already gained all the functionalities like removing the repo from the repositories.json file.

henriquemoraeszup: I think the placement of the delete action makes sense, because to build the tree we **must** have the repo structure in place

henriquemoraeszup (approved):
I'll be damned, 5 lines of code that generated 257 lines of test. Well done sir 👏

pkg/formula/repo/adder.go
```diff
@@ -27,19 +28,22 @@ import (
 type AddManager struct {
 	ritHome string
 	repo    formula.RepositoryListWriteCreator
+	deleter formula.RepositoryDeleter
```
kaduartur, line 31: an improvement for the next implementations would be to have a `formula.RepositoryListWriteCreateDeleter` interface, so you can remove the `deleter` field from your structure and directly use the `repo` parameter

kaduartur (approved):
LGTM!

PR #709, opened 2020-12-04
Deprecating stdin

henriquemoraeszup:
Signed-off-by: Henrique Moraes <henrique.moraes@zup.com.br>

### Description

The team has decided that stdin inputs will be deprecated in favor of flags

### How to verify it
Stdin commands should issue a warning

### Changelog

Deprecated stdin command

### DIFF
--- a/functional/runner.go
+++ b/functional/runner.go
@@ -27,6 +27,8 @@ import (
 	"runtime"
 
 	"github.com/onsi/ginkgo"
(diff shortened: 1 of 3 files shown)


brunasilvazup (approved):
I think an improvement would be to use a single constant for the entire project, but as there are few changes I don't think it is necessary.

PR #768, opened 2020-12-09
Performance new tree struct

kaduartur:
### Description
Create a new command tree format to improve tree generation and merge performance.

We improved all the methods that manipulate the command trees in some way:

- MergedTree has simplified its performance a little.
- GenerateTree we started using `strings.Builder` to improve your performance.
- We created Benchmark tests to visualize the improvements.
- To maintain compatibility, we created a tree converter inside the Ritchie that will be removed in the next versions.

### How to verify it
These changes must not affect or modify the behaviors that already exist in Ritchie-CLI.

### Changelog
Create a new command tree format to improve performance.


pkg/cmd/root.go
```diff
@@ -125,6 +134,10 @@ func (ro *rootCmd) PreRunFunc() CommandRunnerFunc {
 			os.Exit(0)
 		}
 
+		if err := ro.convertTree(); err != nil {
```
henriquemoraeszup, line 137: how do you know a tree is already converted and you don't need to execute it again?

kaduartur: I added a new field called `TreeVersion` in the repositories.json if this field is empty or different from the constant `tree.Version` I execute the update of the command tree to the current repo.

pkg/cmd/add_repo_test.go
```diff
@@ -382,7 +382,7 @@ func TestAddRepoCmd(t *testing.T) {
 
 			err := cmd.Execute()
 
-			assert.Equal(t, tt.wantErr, (err != nil))
+			assert.Equal(t, tt.wantErr, err != nil)
```
henriquemoraeszup, line 385: we dont need to fix this here, but it would be best to check for the error message than the error existence. Through existence, we do not know for sure if the correct error was triggered

pkg/cmd/root.go
```diff
@@ -242,3 +255,55 @@ func (ro *rootCmd) ritchieIsInitialized() bool {
 	commonsRepoPath := filepath.Join(ro.ritchieHome, "repos", "commons")
 	return ro.dir.Exists(commonsRepoPath)
 }
+
+func (ro *rootCmd) convertTree() error {
```
henriquemoraeszup, line 259: should we have a todo to remove this method in the future? We are still opening all trees with this

pkg/cmd/root_test.go
```diff
@@ -219,3 +223,189 @@ func TestConvertContextToEnv(t *testing.T) {
...
+	writeErr error
+}
+
+func (r RepoListWriterMock) List() (formula.Repos, error) {
+	return r.repos, r.listErr
+}
+
+func (r RepoListWriterMock) Write(repos formula.Repos) error {
+	return r.writeErr
+}
```
henriquemoraeszup, line 411: I think we should use mocks from the `mocks` package. With testify mocks, I think this file is good to cross out of the list of maintenance tests

pkg/formula/tree/default_tree_test.go
```diff
@@ -20,407 +20,485 @@ import (
...
+			name: "return empty tree when invalid tree",
+			in: in{
+				repo: repositoryListerCustomMock{
+					list: func() (formula.Repos, error) {
+						return formula.Repos{repoInvalid}, nil
 					},
 				},
+				file:      fileManager,
+				providers: providers,
+				core:      false,
```
henriquemoraeszup, line 106: we could use testify mocks here as well

pkg/api/api.go
```diff
@@ -29,58 +29,77 @@ const (
...
-	Id       string `json:"id"`
 	Parent   string `json:"parent"`
 	Usage    string `json:"usage"`
-	Help     string `json:"help"`
-	LongHelp string `json:"longHelp"`
+	Help     string `json:"help,omitempty"`
+	LongHelp string `json:"longHelp,omitempty"`
 	Formula  bool   `json:"formula,omitempty"`
-	Repo     string `json:"Repo,omitempty"`
+	Repo     string `json:"-"`
```
brunasilvazup, line 99: This is correct?

pkg/formula/tree.go
```diff
@@ -19,11 +19,13 @@ package formula
 import "github.com/ZupIT/ritchie-cli/pkg/api"
 
 type Tree struct {
-	Commands api.Commands `json:"commands"`
+	Version    string          `json:"version"`
+	Commands   api.Commands    `json:"commands"`
+	CommandsID []api.CommandID `json:"-"`
```
brunasilvazup, line 24: Does this `json:"-"` mean that there is no json entry for this field? Could it be removed?

pkg/formula/tree/checker_test.go
```diff
@@ -1,82 +1,127 @@
...
+					Usage:   "vpc",
+					Help:    "create vpc for aws",
+					Formula: true,
+				},
+				"root_aws_create_bucket": {
+					Parent:  "root",
+					Usage:   "bucket",
+					Help:    "create bucket for aws",
+					Formula: true,
+				},
```
brunasilvazup, line 76: If you want you can put these two commands in variables

pkg/formula/tree/default_tree_test.go
```diff
@@ -20,407 +20,485 @@ import (
...
+	_ = os.MkdirAll(filepath.Dir(repo1Path), os.ModePerm)
+	_ = os.MkdirAll(filepath.Dir(repo2Path), os.ModePerm)
+	_ = os.MkdirAll(filepath.Dir(repo3Path), os.ModePerm)
 
-func isSameTree(tree, expected map[string]formula.Tree) bool {
-	if len(tree) != len(expected) {
-		return false
+	_ = fileManager.Write(repo1Path, tree1)
+	_ = fileManager.Write(repo2Path, tree2)
+	_ = fileManager.Write(repo3Path, []byte("invalid"))
```
brunasilvazup, line 327: This config is same in other tests ("TestTree" and "TestMergedTree")?
Could you encapsulate in an auxiliary function?

pkg/formula/tree/default_tree_test.go
```diff
@@ -20,407 +20,485 @@ import (
...
-			Commands: api.Commands{
-				someRepoTree.Commands[0],
-				someRepoTree.Commands[1],
-			},
-		},
-	}
+	_ = fileManager.Write(repo1Path, tree1)
+	_ = fileManager.Write(repo2Path, tree2)
+	_ = fileManager.Write(repo3Path, []byte("invalid"))
+	// End setup
```
brunasilvazup, line 55: This config is same in other tests?
Could be you encapsulate in a auxiliary function?

pkg/formula/tree/generator_test.go
```diff
@@ -45,10 +44,33 @@ func TestGenerate(t *testing.T) {
...
+	_ = dirManager.Remove(workspacePath)
+	_ = dirManager.Create(workspacePath)
+	_ = streams.Unzip("../../../testdata/ritchie-formulas-test.zip", workspacePath)
+
+	b.ResetTimer()
+
+	for i := 0; i < b.N; i++ {
+		_, _ = generator.Generate(workspacePath)
+	}
 }
```
brunasilvazup, line 76: Should that stay?

pkg/api/api.go
kaduartur: yes, it ignores the field when encoding for JSON.

pkg/formula/tree.go
kaduartur: it ignores the field when encoding for JSON, if we remove it we will have a field called CommandsID in the tree.

pkg/formula/tree/generator_test.go
kaduartur: yes, we can measure the performance of the method if someone wants to perform a refactoring

pkg/cmd/root.go
kaduartur: we don't open all the trees, just the ones with the old version

henriquemoraeszup: I thought `TreeVersion` came from the tree file, now I got it. Sounds good

pkg/formula/tree/generator_test.go
henriquemoraeszup: I like this documented in the code as well

henriquemoraeszup (dismissed):
💯

brunasilvazup (dismissed):

pkg/cmd/root_test.go
```diff
@@ -219,3 +226,200 @@ func TestConvertContextToEnv(t *testing.T) {
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			in := tt.in
+			repoMock := new(mocks.RepoManager)
+			repoMock.On("List").Return(in.repo.repos, in.repo.listErr)
+			repoMock.On("Write", mock.Anything).Return(in.repo.writeErr)
+			fileMock := new(mocks.FileManager)
+			fileMock.On("Write", mock.Anything, mock.Anything).Return(in.file.writeErr)
+			treeMock := new(mocks.TreeManager)
```
henriquemoraeszup, line 412: not sure if we should fix this now or later, but I think its best to not mock a 1 line dependency from file manager and actually use `ioutil` in code

henriquemoraeszup (approved):

brunasilvazup (approved):

PR #769, opened 2020-12-09
feat: add multiselect input

### LINKED ISSUE #600: Add a multiselect input type for formulas

lucasdittrichzup:
### Description
This PR adds the type of multiselect input. Thus, from this type of input it will be possible to select more than one item when necessary.

<p align="center">
</p>

In addition, tests with lib testify were added and other tests in the prompt_test.go file were refactored to use tetify as well.

Close #600 

### How to verify it
Create a formula that contains a multiselect input and contains the items field with one or more items

### Changelog
Add multiselect input


pkg/formula/input/prompt/prompt.go
```diff
@@ -138,6 +145,16 @@ func (in InputManager) inputTypeToPrompt(items []string, i formula.Input) (strin
 			return "", err
 		}
 		return in.List(i.Label, dl, i.Tutorial)
+	case input.Multiselect:
```
henriquemoraeszup, line 148: If you added newline between the other `cases`, you should add here as well to follow the same pattern

pkg/formula/input/prompt/prompt.go
```diff
@@ -138,6 +145,16 @@ func (in InputManager) inputTypeToPrompt(items []string, i formula.Input) (strin
...
+		if len(items) > 0 {
+			sl, err := in.Multiselect(i)
+			if err != nil {
+				return "", nil
+			}
+			return strings.Join(sl, ", "), nil
+		}
+		return "", fmt.Errorf(EmptyItems, i.Name)
```
henriquemoraeszup, line 156: I would invert the `if` to make it easier to read, what do you think?
```
if len(items) == 0 {
   return "", fmt.Errorf(EmptyItems, i.Name)
} 
// rest of code
```

pkg/formula/input/prompt/prompt_test.go
```diff
@@ -490,19 +507,24 @@ func TestInputManager_RegexType(t *testing.T) {
...
 			got := inputManager.Inputs(cmd, setup, nil)
 
-			if tt.want != nil && got == nil {
-				t.Errorf("Inputs regex(%s): got %v, want %v", tt.name, nil, tt.want)
-			}
-
-			if tt.want == nil && got != nil {
-				t.Errorf("Inputs regex(%s): got %v, want %v", tt.name, got, nil)
-			}
+			assert.Equal(t, tt.want, got)
```
henriquemoraeszup, line 527: Soooooo much better 🤩

pkg/formula/input/prompt/prompt.go
lucasdittrichzup: Done!

pkg/formula/input/prompt/prompt.go
lucasdittrichzup: Done!

pkg/formula/input/prompt/prompt_test.go
lucasdittrichzup: yes, much clearer to read and easy to test :heart:

pkg/formula/input/prompt/prompt.go
```diff
@@ -131,13 +135,21 @@ func (in InputManager) inputTypeToPrompt(items []string, i formula.Input) (strin
...
 		}
 		return in.List(i.Label, dl, i.Tutorial)
+	case input.Multiselect:
+		if len(items) == 0 {
+			return "", fmt.Errorf(EmptyItems, i.Name)
+		}
+		sl, err := in.Multiselect(i)
+		if err != nil {
+			return "", nil
+		}
```
henriquemoraeszup, line 151: why are you not returning the error?

pkg/formula/input/prompt/prompt.go
```diff
@@ -220,7 +232,7 @@ func (in InputManager) loadItems(input formula.Input, formulaPath string) ([]str
 			var items []string
 			err = json.Unmarshal(fileBytes, &items)
 			if err != nil {
-				return nil, err
+				return nil, errors.New(err.Error())
```
henriquemoraeszup, line 235: why are you instantiating a new error with the content of the current error? We should not complicate the production code just to satisfy tests. Let the complexity stay in the test suite

pkg/prompt/multiselect.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+	multiQs := []*survey.Question{
+		{
+			Prompt: multiselect,
+		},
+	}
+
+	if len(helper) > 0 {
+		multiselect.Help = helper
+	}
```
henriquemoraeszup, line 45: if `helper` is an empty string already, there is no need for the `if`

pkg/prompt/multiselect.go
```diff
@@ -0,0 +1,56 @@
...
+
+	if *input.Required {
+		multiQs[0].Validate = survey.Required
+	}
```
henriquemoraeszup, line 49: is it possible to do this in one line? By always setting `Validate` to the `Required` value

pkg/formula/input/prompt/prompt.go
lucasdittrichzup: my mistake :facepalm:

pkg/prompt/multiselect.go
lucasdittrichzup: not because `Validate` expects a **function / method**, while `Required` is a **boolean**. So, I need to check if the input is mandatory so that I can assign the validation function of the survey itself

pkg/formula/input/prompt/prompt.go
```diff
@@ -131,13 +135,21 @@ func (in InputManager) inputTypeToPrompt(items []string, i formula.Input) (strin
...
 		return in.List(i.Label, dl, i.Tutorial)
+	case input.Multiselect:
+		if len(items) == 0 {
+			return "", fmt.Errorf(EmptyItems, i.Name)
+		}
+		sl, err := in.Multiselect(i)
+		if err != nil {
+			return "", err
+		}
+		return strings.Join(sl, ", "), nil
```
henriquemoraeszup, line 152: hey, on the test, the join with space became a little odd. The user will have to know to split with the comma and space to get the correct item. Also, if the item has a comma by itself, it might mess up the split, like:
What's your favorite food:
[] fish, shrimp, seafood in general
[] pasta
[] meat

Given this, what about joining with pipe `|`?

brunasilvazup (dismissed):

pkg/formula/input/prompt/prompt.go
lucasdittrichzup: Done!

henriquemoraeszup (approved):

PR #774, opened 2020-12-14
Delete credential flags

henriquemoraeszup:
### Description

Added flag capacity on delete credential

### How to verify it
Run the command `rit delete credential --provider=<your provider>`

### Changelog

Added flag capacity to `rit delete credential`


pkg/cmd/delete_credential.go
```diff
@@ -97,19 +90,64 @@ func (d deleteCredentialCmd) runPrompt() CommandRunnerFunc {
...
+	if len(data) == 0 {
+		return inputConfig{}, errors.New("you have no defined credentials in this env")
+	}
+
+	providers := make([]string, len(data))
+	for _, c := range data {
+		providers = append(providers, c.Provider)
+	}
+
+	cred, err := d.List("Credentials: ", providers)
```
brunasilvazup, line 116: This name of variable could be change `provider` to equal in line 130?

brunasilvazup (approved):

pkg/cmd/delete_credential.go
```diff
@@ -97,19 +90,64 @@ func (d deleteCredentialCmd) runPrompt() CommandRunnerFunc {
 	}
 }
 
+func (d *deleteCredentialCmd) resolveInput(cmd *cobra.Command, context string) (inputConfig, error) {
+	if IsFlagInput(cmd) {
+		return d.resolveFlags(cmd)
+	} else {
+		return d.resolvePrompt(context)
+	}
```
brunasilvazup, line 98: This `else` is not necessary, right?
```go
} else {
		return d.resolvePrompt(context)
}
```
->
```go
}
return d.resolvePrompt(context)
```

brunasilvazup (approved):

pkg/cmd/cmd.go
```diff
@@ -45,6 +45,10 @@ func RunFuncE(stdinFunc, promptFunc CommandRunnerFunc) CommandRunnerFunc {
 	}
 }
 
+func IsFlagInput(cmd *cobra.Command) bool {
+	return cmd.Flags().NFlag() > 0
```
kaduartur, line 49: When you use the `--help` flag, does it work?

pkg/cmd/delete_credential.go
```diff
@@ -12,6 +13,15 @@ import (
 	"github.com/ZupIT/ritchie-cli/pkg/stdin"
 )
 
+const (
+	providerFlagName        = "provider"
```
kaduartur, line 17: could you extract the `flag` struct from the formula.go and use `addReservedFlags` function to generate all flags to any commands?

pkg/cmd/delete_credential.go
```diff
@@ -46,49 +56,32 @@ func NewDeleteCredentialCmd(
...
 
-		data, err := d.ReadCredentialsValueInEnv(d.CredentialsPath(), env)
+		inputParams, err := d.resolveInput(cmd, curEnv)
 		if err != nil {
 			return err
-		}
-
-		if len(data) <= 0 {
-			prompt.Error("You have no defined credentials in this env")
+		} else if inputParams.provider == "" {
```
kaduartur, line 80: I think you should return an error when the user sends an empty provider, so you don't need to check if the provider is empty here.

pkg/cmd/delete_credential.go
```diff
@@ -97,19 +90,64 @@ func (d deleteCredentialCmd) runPrompt() CommandRunnerFunc {
...
+	}
+	return inputConfig{provider}, nil
+}
+
+func (d *deleteCredentialCmd) resolveFlags(cmd *cobra.Command) (inputConfig, error) {
+	provider, err := cmd.Flags().GetString(providerFlagName)
+	if err != nil {
+		return inputConfig{}, err
+	}
+	return inputConfig{provider}, nil
```
kaduartur, line 134: you should return an error if the provider is empty

kaduartur (changes requested):

pkg/cmd/cmd.go
henriquemoraeszup: good point, but it works as intended, seems like `help` is a cobra reserved flag that gets priority when called

pkg/cmd/delete_credential.go
henriquemoraeszup: Sorry I am a bit lost, are core commands supposed to have reserved flags?

pkg/cmd/delete_credential.go
henriquemoraeszup: Returning the error makes sense, but we cannot neglect this case. This happens when we prompt the user if she really wants to delete and she says no

pkg/cmd/delete_credential.go
```diff
@@ -46,49 +57,36 @@ func NewDeleteCredentialCmd(
...
 
-func (d deleteCredentialCmd) runPrompt() CommandRunnerFunc {
-	return func(cmd *cobra.Command, args []string) error {
-		env, err := d.currentEnv()
-		if err != nil {
-			return err
-		}
-		prompt.Info(fmt.Sprintf("Current env: %s", env))
+func addFlags(flags *pflag.FlagSet) {
+	flags.String(providerFlagName, "", providerFlagDescription)
```
kaduartur, line 71: - I think that you can use the `flag` struct to represent a flag.
https://github.com/ZupIT/ritchie-cli/blob/84b00081d6ffcb6177852478051d090468891eb8/pkg/cmd/formula.go#L78
 
- you can use this function to create your flags.
https://github.com/ZupIT/ritchie-cli/blob/84b00081d6ffcb6177852478051d090468891eb8/pkg/cmd/formula.go#L227

- here we have an example of how to use it:
https://github.com/ZupIT/ritchie-cli/blob/84b00081d6ffcb6177852478051d090468891eb8/pkg/cmd/formula.go#L47

you can extract this function and struct to [cmd.go](https://github.com/ZupIT/ritchie-cli/blob/master/pkg/cmd/cmd.go) file

kaduartur (changes requested):

pkg/cmd/delete_credential.go
henriquemoraeszup: now I see...

brunasilvazup (approved):

kaduartur (approved):
Good job 🚀

pkg/cmd/delete_credential.go
```diff
@@ -97,19 +100,65 @@ func (d deleteCredentialCmd) runPrompt() CommandRunnerFunc {
...
+	data, err := d.ReadCredentialsValueInEnv(d.CredentialsPath(), context)
+	if err != nil {
+		return inputConfig{}, err
+	}
+
+	if len(data) == 0 {
+		return inputConfig{}, errors.New("you have no defined credentials in this env")
+	}
+
+	providers := make([]string, len(data))
```
kaduartur, line 120: When you declare an array like this, you are saying that its capacity is the date size and you are giving each position of the array an empty value, by appending it will add to the next position displaying to the user a select with empty fields.

image


For example, if the date has a size of 3 you will have an array with size 3, and when appending it will add one more position and the size of your array will be 6 with the first 3 empty positions.

we have two ways to fix this:
``` go

var providers []string // Simpler

providers := make([]string, 0, len(data)) // More complex
```

kaduartur (changes requested):

pkg/cmd/delete_credential.go
henriquemoraeszup: the first option does not pass the linter since we need to preallocate it, going for the second

kaduartur (approved):

PR #777, opened 2020-12-14
Fix/struct solution

victorschumacherzup:
### Description

A bug related to tag struct without the correct json key
### How to verify it
use rit add repo


kaduartur (dismissed):

pkg/git/gitlab/repository.go
```diff
@@ -123,21 +123,14 @@ func (re RepoManager) LatestTag(info git.RepoInfo) (git.Tag, error) {
 		return git.Tag{}, errors.New(string(b))
 	}
 
-	var gitlabTag []struct {
-		Name string `json:"tag_name"`
-	}
-
-	if err := json.NewDecoder(res.Body).Decode(&gitlabTag); err != nil {
+	var tags git.Tags
+	if err := json.NewDecoder(res.Body).Decode(&tags); err != nil {
```
brunasilvazup, line 127: wouldn't this be just an element (tags -> tag)?

brunasilvazup (approved):

PR #808, opened 2020-12-19
Tests on slice

henriquemoraeszup:
### Description

There were some changes on this PR besides adding testify. Some of the methods from the utils package were used in a single file. Such methods were moved to their respective files. Also to avoid folder nesting, `slice_util.go` was moved up one folder

### How to verify it
Run slice tests and check if they pass

### Changelog

Test refactor on slice util


brunasilvazup (approved):

kaduartur (changes requested):
Keep the `sliceutil/` package inside the `slice/` so we don't have **"util"** packages lost inside `pkg/`, this is a very common pattern in Golang. An example is the [ioutil](https://golang.org/pkg/io/ioutil) package.

kaduartur (approved):
I liked the approach of removing functions that are specific to their respective packages! 🚀

Issue #802, opened 2020-12-16
Create a RIT ADD WORKSPACE command


GuillaumeFalourd:
**What would you like to be added**:

A core command to add a new workspace.

**Why is this needed**:

Today, you can only add a new workspace when creating a formula.
 
However, there is a scenario when an user may need to update a formula without having the workspace configured locally, which is when cloning a new repository to update a formula inside it.

As it is today, if an user clones a repository to edit a formula inside it, he won't be able to test it locally without adding a new formula to the cloned repository, to turn it a workspace (which is strange).

- [X] Check this box if this issue needs a documentation update

henriquemoraeszup:
Flags added `name` -> workspace name, `path` -> workspace path

PR #809, opened 2020-12-19
created rit add workspace command

aronrichter:
### Description
As requested on #802, is necessary a command do add a new workspace.

### How to verify it
You can run the command `rit add workspace` and after you can run `rit list workspace` and then you can see your workspace added

### Changelog
Added a new core command `rit add workspace`

### DIFF
--- a/pkg/cmd/add_workspace.go
+++ b/pkg/cmd/add_workspace.go
@@ -1,0 +1,103 @@
+package cmd
+
+import (
+	"errors"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/ZupIT/ritchie-cli/pkg/formula"
+	"github.com/ZupIT/ritchie-cli/pkg/prompt"
+)
+
+const (
+	workspaceNameFlag            = "name"
+	workspaceNameFlagDescription = "Workspace name"
+	workspacePathFlag            = "path"
+	workspacePathFlagDescription = "Workspace path"
+)
+
+type addWorkspaceCmd struct {
+	workspace formula.WorkspaceAddLister
+	input     prompt.InputText
+}
+
+func NewAddWorkspaceCmd(
+	workspace formula.WorkspaceAddLister,
+	input prompt.InputText,
+) *cobra.Command {
+	a := addWorkspaceCmd{
+		workspace: workspace,
+		input:     input,
+	}
+
+	cmd := &cobra.Command{
+		Use:       "workspace",
+		Short:     "Add new workspace",
+		Example:   "rit add workspace",
+		RunE:      a.runFormula(),
+		ValidArgs: []string{""},
+		Args:      cobra.OnlyValidArgs,
+	}
+
+	flags := cmd.Flags()
+	flags.String(workspaceNameFlag, "", workspaceNameFlagDescription)
+	flags.String(workspacePathFlag, "", workspacePathFlagDescription)
+
+	return cmd
+}
+
+func (a addWorkspaceCmd) runFormula() CommandRunnerFunc {
+	return func(cmd *cobra.Command, args []string) error {
+		if cmd.Flags().NFlag() == 0 {
+			wspace, err := addWorkspaceFromPrompt(a)
+			if err != nil {
+				return err
+			}
+
+			return a.workspace.Add(wspace)
+		}
+
+		wspace, err := addWorkspaceFromFlags(cmd)
+		if err != nil {
+			return err
+		}
+
+		return a.workspace.Add(wspace)
+	}
+}
+
+func addWorkspaceFromFlags(cmd *cobra.Command) (formula.Workspace, error) {
+	workspaceName, _ := cmd.Flags().GetString(workspaceNameFlag)
+	workspacePath, _ := cmd.Flags().GetString(workspacePathFlag)
+
+	if len(workspaceName) == 0 || len(workspacePath) == 0 {
+		return formula.Workspace{}, errors.New("all flags need to be filled")
+	}
+
+	wspace := formula.Workspace{
+		Name: strings.Title(workspaceName),
+		Dir:  workspacePath,
+	}
+
+	return wspace, nil
+}
+
+func addWorkspaceFromPrompt(a addWorkspaceCmd) (formula.Workspace, error) {
+	workspaceName, err := a.input.Text("Enter the name of workspace", true)
+	if err != nil {
+		return formula.Workspace{}, err
+	}
+
+	workspacePath, err := a.input.Text("Enter the path of workspace (e.g.: /home/user/github)", true)
+	if err != nil {
+		return formula.Workspace{}, err
+	}
+
+	wspace := formula.Workspace{
+		Name: strings.Title(workspaceName),
+		Dir:  workspacePath,
+	}
+
+	return wspace, nil
+}
--- a/pkg/cmd/add_workspace_test.go
+++ b/pkg/cmd/add_workspace_test.go
@@ -1,0 +1,83 @@
+package cmd
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+
+	"github.com/ZupIT/ritchie-cli/internal/mocks"
+	"github.com/ZupIT/ritchie-cli/pkg/formula"
+)
+
+func TestNewAddWorkspaceCmd(t *testing.T) {
+	var tests = []struct {
+		name                  string
+		addWorkspaceWithError error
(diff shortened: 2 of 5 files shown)


pkg/cmd/add_workspace.go
```diff
@@ -0,0 +1,61 @@
...
+
+	cmd := &cobra.Command{
+		Use:       "workspace",
+		Short:     "Add new workspace",
+		Example:   "rit add workspace",
+		RunE:      a.runPrompt(),
+		ValidArgs: []string{""},
+		Args:      cobra.OnlyValidArgs,
+	}
+
```
henriquemoraeszup, line 34: Hey! we just started adding support for flags on core commands, can you add here as well? you can check #774 for a reference =]

pkg/cmd/add_workspace_test.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			addNewWorkspace := NewAddWorkspaceCmd(
+				tt.wspace,
+				tt.it,
+			)
+			if err := addNewWorkspace.Execute(); (err != nil) != tt.wantErr {
+				t.Errorf("%s = %v, want %v", addNewWorkspace.Use, err, nil)
+			}
```
henriquemoraeszup, line 57: Can you use testify assertions and mocks here? We are adopting a new lib to make our tests clearer. You can find an example here
https://github.com/ZupIT/ritchie-cli/pull/774/files#diff-d694f418a29ad6844afed8da9f20f717d0d953966216b43c3e1155d409f08779R363

henriquemoraeszup (commented):
Thanks for your contribution aron! Just a couple of points below, keep it up 💪

pkg/cmd/add_workspace.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+}
+
+func hasFlags(cmd *cobra.Command) (bool, formula.Workspace) {
+	workspaceName, _ := cmd.Flags().GetString(workspaceNameFlag)
+	workspacePath, _ := cmd.Flags().GetString(workspacePathFlag)
+
+	if len(workspaceName) == 0 || len(workspacePath) == 0 {
+		return false, formula.Workspace{}
+	}
```
henriquemoraeszup, line 70: I think we should return an error if the user has any flags but forgets to fill them instead of falling back to prompt

pkg/cmd/add_workspace.go
```diff
@@ -0,0 +1,97 @@
...
+
+	flags := cmd.Flags()
+	flags.String(workspaceNameFlag, "", workspaceNameFlagDescription)
+	flags.String(workspacePathFlag, "", workspacePathFlagDescription)
+
+	return cmd
+}
+
+func (a addWorkspaceCmd) runPrompt() CommandRunnerFunc {
```
henriquemoraeszup, line 49: nitpick: you can change the method name to something more semantic like `runFormula` because we are not necessarily running a prompt now

pkg/cmd/add_workspace_test.go
```diff
@@ -0,0 +1,59 @@
...
+		argsName              string
+		argsPath              string
+		wantErr               string
+	}{
+		{
+			name:                  "add new workspace by prompt",
+			addWorkspaceWithError: nil,
+			wantErr:               "",
+		},
+		{
```
henriquemoraeszup, line 26: can we have an error case for a user that only adds one flag?

pkg/cmd/add_workspace.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+}
+
+func addWorkspaceFromFlags(cmd *cobra.Command) (formula.Workspace, error, bool) {
+	workspaceName, _ := cmd.Flags().GetString(workspaceNameFlag)
+	workspacePath, _ := cmd.Flags().GetString(workspacePathFlag)
+
+	if len(workspaceName) == 0 && len(workspacePath) == 0 {
+		return formula.Workspace{}, nil, false
+	}
```
henriquemoraeszup, line 76: what happens if you do `rit add workspace --name= --path=`?

pkg/cmd/add_workspace_test.go
```diff
@@ -0,0 +1,66 @@
...
+func TestNewAddWorkspaceCmd(t *testing.T) {
+	var tests = []struct {
+		name                  string
+		addWorkspaceWithError error
+		argsName              string
+		argsPath              string
+		wantErr               string
+	}{
+		{
+			name:                  "add new workspace by prompt",
```
henriquemoraeszup, line 22: One detail, on a success case, you want to check if the workspace entry was actually added to `~/.rit/formula_workspaces.json` so maybe you do not want to mock the workspace added

pkg/cmd/add_workspace.go
aronrichter: It was the same thing when you don't pass any flag.
But, I understood why you'd ask, so I changed to check the number of flags passed before I decided what to do.

pkg/cmd/add_workspace_test.go
aronrichter: I'm not sure if I wanted to check that. 
My unit it's until I called the method `a.workspace.Add(wspace)` after that the behavior ins't responsability of this method. Because of this I mocked my test. 
What do you think?!

henriquemoraeszup: I do agree with you, lately we have been trying to adopt a more functional-like approach on our tests. But your argument is good as well, if you intend to continue with the mock, then maybe it is best to test that it has been called with the right arguments to ensure a thorough test. Check out assertCalled
https://godoc.org/github.com/stretchr/testify/mock#Mock.AssertCalled

henriquemoraeszup (commented):
Just one last suggestion on the mock but I am good to go. Just not approving now because any further commit would dismiss my review

pkg/cmd/add_workspace_test.go
aronrichter: That's right, I can mock my test, but at least a need to guarantee I'm passing the right parameters.
Thanks for the observation! 😄

pkg/cmd/add_workspace_test.go
```diff
@@ -0,0 +1,83 @@
...
+			textMock.On("Text", "Enter the name of workspace", true, mock.Anything).Return(tt.workspaceName, nil)
+			textMock.On("Text", "Enter the path of workspace (e.g.: /home/user/github)", true, mock.Anything).Return(tt.workspacePath, nil)
+
+			wspace := formula.Workspace{
+				Name: tt.workspaceName,
+				Dir:  tt.workspacePath,
+			}
+
+			workspaceMock := &mocks.WorkspaceMock{}
+			workspaceMock.On("Add", wspace).Return(tt.addWorkspaceWithError)
```
henriquemoraeszup, line 70: nice, this is even better than what I suggested 😄 will be recommending this to the team

henriquemoraeszup (approved):
👏 👏 👏 👏 
giphy (12)

pkg/cmd/add_workspace.go
```diff
@@ -0,0 +1,103 @@
...
+const (
+	workspaceNameFlag            = "name"
+	workspaceNameFlagDescription = "Workspace name"
+	workspacePathFlag            = "path"
+	workspacePathFlagDescription = "Workspace path"
+)
+
+type addWorkspaceCmd struct {
+	workspace formula.WorkspaceAddLister
+	inText    prompt.InputText
```
kaduartur, line 22: could you rename this to `input` to make it more readable?

pkg/cmd/add_workspace.go
```diff
@@ -0,0 +1,103 @@
...
+	flags := cmd.Flags()
+	flags.String(workspaceNameFlag, "", workspaceNameFlagDescription)
+	flags.String(workspacePathFlag, "", workspacePathFlagDescription)
+
+	return cmd
+}
+
+func (a addWorkspaceCmd) runFormula() CommandRunnerFunc {
+	return func(cmd *cobra.Command, args []string) error {
+		if count := cmd.Flags().NFlag(); count == 0 {
```
kaduartur, line 52: could you simplify it to `cmd.Flags().NFlag() == 0`?

kaduartur (changes requested):
Good job 🚀 

You just needed to add this command to the core command list, you can add it in the [api.go](https://github.com/ZupIT/ritchie-cli/blob/master/pkg/api/api.go) file. we use this to generate our autocomplete of the core commands.

kaduartur (approved):
LGTM 🥇

PR #823, opened 2021-01-08
feat: add bitbucket as a provider to add repos

### LINKED ISSUE #647: Add Bitbucket to provider list for rit add repo
## What would you like to see

- I would like to be able to import a bitbucket repo on Ritchie through the `rit add repo` command.

## Why is it needed

- Bitbucket is not supported yet, only Github and Gitlab.

lucasdittrichzup:
### Description
This PR adds Bitbucket as a provider to add repos.

<p align="center">
</p>

Close #647 

### How to verify it
- Add new repo with rit add repo
- Run the unit tests with make unit-test command

### Changelog
Add Bitbucket provider to add repos

### DIFF
--- a/internal/mocks/mocks.go
+++ b/internal/mocks/mocks.go
@@ -381,32 +381,3 @@ func (w *WorkspaceMock) List() (formula.Workspaces, error) {
 	args := w.Called()
 	return args.Get(0).(formula.Workspaces), args.Error(1)
 }
-
-type RepoInfo struct {
-	mock.Mock
-}
-
-func (ri *RepoInfo) ZipUrl(version string) string {
-	args := ri.Called(version)
-	return args.String(0)
-}
-
-func (ri *RepoInfo) TagsUrl() string {
-	args := ri.Called()
-	return args.String(0)
-}
-
-func (ri *RepoInfo) LatestTagUrl() string {
-	args := ri.Called()
-	return args.String(0)
-}
-
-func (ri *RepoInfo) TokenHeader() string {
-	args := ri.Called()
-	return args.String(0)
-}
-
-func (ri *RepoInfo) Token() string {
-	args := ri.Called()
-	return args.String(0)
-}
--- a/internal/mocks/mocks.go
+++ b/internal/mocks/mocks.go
@@ -253,6 +253,20 @@ func (r *RepoManager) LatestTag(repo formula.Repo) string {
 	return args.String(0)
 }
 
+type DirManager struct {
+	mock.Mock
+}
+
+func (d *DirManager) Exists(path string) bool {
+	args := d.Called(path)
+	return args.Bool(0)
+}
+
+func (d *DirManager) IsDir(dir string) bool {
+	args := d.Called(dir)
+	return args.Bool(0)
+}
+
 type FileManager struct {
 	mock.Mock
 }
--- a/internal/mocks/mocks.go
+++ b/internal/mocks/mocks.go
@@ -93,7 +93,6 @@ type InputURLMock struct {
 
 func (i *InputURLMock) URL(name, defaultValue string) (string, error) {
 	args := i.Called(name, defaultValue)
-
 	return args.String(0), args.Error(1)
 }
 
--- a/internal/mocks/mocks.go
+++ b/internal/mocks/mocks.go
@@ -381,3 +381,32 @@ func (w *WorkspaceMock) List() (formula.Workspaces, error) {
 	args := w.Called()
 	return args.Get(0).(formula.Workspaces), args.Error(1)
 }
+
+type RepoInfo struct {
+	mock.Mock
+}
+
+func (ri *RepoInfo) ZipUrl(version string) string {
+	args := ri.Called(version)
+	return args.String(0)
+}
+
+func (ri *RepoInfo) TagsUrl() string {
+	args := ri.Called()
+	return args.String(0)
+}
+
+func (ri *RepoInfo) LatestTagUrl() string {
+	args := ri.Called()
+	return args.String(0)
+}
+
+func (ri *RepoInfo) TokenHeader() string {
+	args := ri.Called()
+	return args.String(0)
+}
+
+func (ri *RepoInfo) Token() string {
+	args := ri.Called()
+	return args.String(0)
+}
--- a/pkg/git/bitbucket/bitbucket.go
+++ b/pkg/git/bitbucket/bitbucket.go
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2020 ZUP IT SERVICOS EM TECNOLOGIA E INOVACAO SA
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package bitbucket
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/ZupIT/ritchie-cli/pkg/git"
+)
+
+const (
+	ZipUrlPattern       = "https://bitbucket.org/%s/%s/get/%s.zip"
+	TagsUrlPattern      = "https://api.bitbucket.org/2.0/repositories/%s/%s/refs/tags"
+	LatestTagUrlPattern = "https://api.bitbucket.org/2.0/repositories/%s/%s/refs/tags?sort=target.date"
+)
+
+type DefaultRepoInfo struct {
+	host  string
+	owner string
+	repo  string
+	token string
+}
+
+// NewRepoInfo returns the RepoInfo built by repository url
+// Repository url e.g. https://bitbucket.org/{{owner}}/{{repo}}/src/master/
+func NewRepoInfo(url string, token string) git.RepoInfo {
+	split := strings.Split(url, "/")
+	if len(split) < 8 {
+		return DefaultRepoInfo{}
+	}
+
+	repo := split[len(split)-4]
+	owner := split[len(split)-5]
+	host := split[len(split)-6]
+
+	return DefaultRepoInfo{
+		host:  host,
+		owner: owner,
+		repo:  repo,
+		token: token,
+	}
+}
+
+// ZipUrl returns the Bitbucket API URL for download zipball repository
+// e.g. https://bitbucket.org/{{owner}}/{{repo}}/get/{{tag-version}}.zip
+func (in DefaultRepoInfo) ZipUrl(version string) string {
+	return fmt.Sprintf(ZipUrlPattern, in.owner, in.repo, version)
+}
+
+// TagsUrl returns the Bitbucket API URL for get all tags
+// e.g. https://api.bitbucket.org/2.0/repositories/{{owner}}/{{repo}}/refs/tags
+func (in DefaultRepoInfo) TagsUrl() string {
+	return fmt.Sprintf(TagsUrlPattern, in.owner, in.repo)
+}
+
+// LatestTagUrl returns the Bitbucket API URL for get latest tag release
+// e.g. https://api.bitbucket.org/2.0/repositories/{{owner}}/{{repo}}/refs/tags?sort=target.date
+func (in DefaultRepoInfo) LatestTagUrl() string {
+	return fmt.Sprintf(LatestTagUrlPattern, in.owner, in.repo)
+}
+
+// TokenHeader returns the Authorization value formatted for Bitbucket API integration
+// e.g. "m6ioSP4o4q6tmiinXHf9KOHxQsjbvShS-zMullcRsxiJtMHIqr2tuHpJZbl-UXpm7E-1meNlAqKzmORTxyoNAnXWZlCFPsvJpQj4evGtafuH4NBBgRrQ_Mc3"
+func (in DefaultRepoInfo) TokenHeader() string {
+	return in.token
+}
+
+func (in DefaultRepoInfo) Token() string {
+	return in.token
+}
--- a/pkg/git/bitbucket/bitbucket.go
+++ b/pkg/git/bitbucket/bitbucket.go
@@ -39,13 +39,14 @@ const (
 	repoOffset  = 4
 	ownerOffset = 5
 	hostOffset  = 6
+	defaultUrl  = 8
 )
 
 // NewRepoInfo returns the RepoInfo built by repository url
 // Repository url e.g. https://bitbucket.org/{{owner}}/{{repo}}/src/master/
 func NewRepoInfo(url string, token string) git.RepoInfo {
 	split := strings.Split(url, "/")
-	if len(split) < 8 {
+	if len(split) < defaultUrl {
 		return DefaultRepoInfo{}
 	}
 
--- a/pkg/git/bitbucket/bitbucket.go
+++ b/pkg/git/bitbucket/bitbucket.go
@@ -35,6 +35,12 @@ type DefaultRepoInfo struct {
 	token string
 }
 
+const (
+	repoOffset  = 4
+	ownerOffset = 5
+	hostOffset  = 6
+)
+
 // NewRepoInfo returns the RepoInfo built by repository url
 // Repository url e.g. https://bitbucket.org/{{owner}}/{{repo}}/src/master/
 func NewRepoInfo(url string, token string) git.RepoInfo {
@@ -43,9 +49,9 @@ func NewRepoInfo(url string, token string) git.RepoInfo {
 		return DefaultRepoInfo{}
 	}
 
-	repo := split[len(split)-4]
-	owner := split[len(split)-5]
-	host := split[len(split)-6]
+	repo := split[len(split)-repoOffset]
+	owner := split[len(split)-ownerOffset]
+	host := split[len(split)-hostOffset]
 
 	return DefaultRepoInfo{
 		host:  host,
@@ -74,7 +80,6 @@ func (in DefaultRepoInfo) LatestTagUrl() string {
 }
 
 // TokenHeader returns the Authorization value formatted for Bitbucket API integration
-// e.g. "m6ioSP4o4q6tmiinXHf9KOHxQsjbvShS-zMullcRsxiJtMHIqr2tuHpJZbl-UXpm7E-1meNlAqKzmORTxyoNAnXWZlCFPsvJpQj4evGtafuH4NBBgRrQ_Mc3"
 func (in DefaultRepoInfo) TokenHeader() string {
 	return in.token
 }
--- a/pkg/git/bitbucket/bitbucket_test.go
+++ b/pkg/git/bitbucket/bitbucket_test.go
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2020 ZUP IT SERVICOS EM TECNOLOGIA E INOVACAO SA
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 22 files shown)


pkg/git/bitbucket/repository_test.go
```diff
@@ -0,0 +1,470 @@
...
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			re := NewRepoManager(tt.in.client)
+
+			got, err := re.Tags(tt.in.info)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("Tags() error = %v, wantErr %v", err, tt.wantErr)
```
brunasilvazup, line 83: Could you use testify asserts here?

lucasdittrichzup: i will, but I decided to open the PR for review before

pkg/git/bitbucket/bitbucket.go
```diff
@@ -0,0 +1,84 @@
...
+	repo  string
+	token string
+}
+
+// NewRepoInfo returns the RepoInfo built by repository url
+// Repository url e.g. https://bitbucket.org/{{owner}}/{{repo}}/src/master/
+func NewRepoInfo(url string, token string) git.RepoInfo {
+	split := strings.Split(url, "/")
+	if len(split) < 8 {
+		return DefaultRepoInfo{}
```
henriquemoraeszup, line 43: if someone provides the wrong url, should this be treated as an error? Can Bitbucket users end up passing just `https://bitbucket.org/{{owner}}/{{repo}}`? We should consider user experience here

pkg/git/bitbucket/bitbucket.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	repo := split[len(split)-4]
+	owner := split[len(split)-5]
+	host := split[len(split)-6]
```
henriquemoraeszup, line 48: Can you make consts for these magic numbers? like `REPO_OFFSET`, `OWNER_OFFSET` etc

https://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad

pkg/git/bitbucket/bitbucket.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// LatestTagUrl returns the Bitbucket API URL for get latest tag release
+// e.g. https://api.bitbucket.org/2.0/repositories/{{owner}}/{{repo}}/refs/tags?sort=target.date
+func (in DefaultRepoInfo) LatestTagUrl() string {
+	return fmt.Sprintf(LatestTagUrlPattern, in.owner, in.repo)
+}
+
+// TokenHeader returns the Authorization value formatted for Bitbucket API integration
+// e.g. "m6ioSP4o4q6tmiinXHf9KOHxQsjbvShS-zMullcRsxiJtMHIqr2tuHpJZbl-UXpm7E-1meNlAqKzmORTxyoNAnXWZlCFPsvJpQj4evGtafuH4NBBgRrQ_Mc3"
```
henriquemoraeszup, line 77: this token is invalid right? I would remove this line just to not raise flags for security

pkg/git/bitbucket/bitbucket_test.go
```diff
@@ -0,0 +1,55 @@
...
+	"reflect"
+	"testing"
+)
+
+func TestNewRepoInfo(t *testing.T) {
+	want := DefaultRepoInfo{
+		host:  "bitbucket.org",
+		owner: "username",
+		repo:  "ritchie-formulas",
+		token: "gHexna7h7CQWafwNYSXp",
```
henriquemoraeszup, line 28: is this token fake? You can provide any string here, since the token itself does not matter, like `some-token`. Also since you are repeating this value in every test, you could make a constant at the top of the file

pkg/git/bitbucket/bitbucket_test.go
```diff
@@ -0,0 +1,55 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package bitbucket
+
+import (
+	"reflect"
+	"testing"
```
henriquemoraeszup, line 20: Remember to use testify

pkg/git/bitbucket/repository.go
```diff
@@ -0,0 +1,159 @@
...
+	Values []value `json:"values"`
+}
+
+func NewRepoManager(client *http.Client) RepoManager {
+	return RepoManager{client: client}
+}
+
+func (re RepoManager) Zipball(info git.RepoInfo, version string) (io.ReadCloser, error) {
+	zipUrl := info.ZipUrl(version)
+	req, err := http.NewRequestWithContext(context.TODO(), http.MethodGet, zipUrl, nil)
```
henriquemoraeszup, line 49: `TODO` context is used if you are unsure about the type of context to use. You can probably use `context.Background()` here

pkg/git/bitbucket/repository.go
```diff
@@ -0,0 +1,159 @@
...
+		authToken := info.TokenHeader()
+		req.Header.Add(headers.Authorization, fmt.Sprintf("Bearer %s", authToken))
+	}
+
+	resp, err := re.client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+
+	if resp.StatusCode != 200 {
```
henriquemoraeszup, line 64: Can the success status code be different from 200? Like a 202 or 201?

pkg/git/bitbucket/bitbucket.go
```diff
@@ -0,0 +1,84 @@
...
+func (in DefaultRepoInfo) TokenHeader() string {
+	return in.token
+}
+
+func (in DefaultRepoInfo) Token() string {
+	return in.token
+}
```
henriquemoraeszup, line 84: These two methods are returning the same thing, why?

pkg/git/bitbucket/bitbucket.go
```diff
@@ -0,0 +1,84 @@
...
+	ZipUrlPattern       = "https://bitbucket.org/%s/%s/get/%s.zip"
+	TagsUrlPattern      = "https://api.bitbucket.org/2.0/repositories/%s/%s/refs/tags"
+	LatestTagUrlPattern = "https://api.bitbucket.org/2.0/repositories/%s/%s/refs/tags?sort=target.date"
+)
+
+type DefaultRepoInfo struct {
+	host  string
+	owner string
+	repo  string
+	token string
```
henriquemoraeszup, line 35: Just to debate here, would it be easier to make these fields public so other files can access it? Or maybe just the token public to not implement the getter?

pkg/git/bitbucket/repository.go
```diff
@@ -0,0 +1,159 @@
...
+
+func (re RepoManager) Tags(info git.RepoInfo) (git.Tags, error) {
+	apiUrl := info.TagsUrl()
+	req, err := http.NewRequestWithContext(context.TODO(), http.MethodGet, apiUrl, nil)
+	if err != nil {
+		return git.Tags{}, err
+	}
+
+	if info.Token() != "" {
+		authToken := info.TokenHeader()
```
henriquemoraeszup, line 83: If the return value of these two methods are the same (`in.token`), how does this logic work?

henriquemoraeszup: Also, if every tag request is authenticated, I would unconditionally set the Authorization header

pkg/git/bitbucket/repository.go
```diff
@@ -0,0 +1,159 @@
...
+
+	res, err := re.client.Do(req)
+	if err != nil {
+		return git.Tags{}, err
+	}
+
+	defer res.Body.Close()
+
+	if res.StatusCode != http.StatusOK {
+		errorMessage := fmt.Sprintf("There was an error adding the repository, status: %d - %s.", res.StatusCode, http.StatusText(res.StatusCode))
```
henriquemoraeszup, line 95: Can we print the specific error message that would come from `res`? Just the status can be too generic for the user to troubleshoot

pkg/git/bitbucket/repository.go
```diff
@@ -0,0 +1,159 @@
...
+	var tags git.Tags
+	for _, v := range bTags.Values {
+		tag := git.Tag{Name: v.Name}
+		tags = append(tags, tag)
+	}
+
+	return tags, nil
+}
+
+func (re RepoManager) LatestTag(info git.RepoInfo) (git.Tag, error) {
```
henriquemoraeszup, line 113: Same comments about tokens and contexts apply here

pkg/git/bitbucket/repository.go
```diff
@@ -0,0 +1,159 @@
...
+	var bTags bitbucketTags
+	if err := json.NewDecoder(res.Body).Decode(&bTags); err != nil {
+		return git.Tag{}, err
+	}
+
+	var tags git.Tags
+	for _, v := range bTags.Values {
+		tag := git.Tag{Name: v.Name}
+		tags = append(tags, tag)
+	}
```
henriquemoraeszup, line 150: If you only need the last, you do not need to iterate all and build the array, just build the last one. Do not forget to handle the case of an empty array

pkg/git/bitbucket/repository_test.go
```diff
@@ -0,0 +1,470 @@
...
+
+	type in struct {
+		client *http.Client
+		info   git.RepoInfo
+	}
+	tests := []struct {
+		name    string
+		in      in
+		want    git.Tags
+		wantErr bool
```
henriquemoraeszup, line 46: We should strive for the specific error, not the existence of one

pkg/git/bitbucket/repository_test.go
```diff
@@ -0,0 +1,470 @@
...
+		writer.WriteHeader(http.StatusBadRequest)
+	}))
+
+	type in struct {
+		client *http.Client
+		info   git.RepoInfo
+	}
+	tests := []struct {
+		name    string
+		in      in
```
henriquemoraeszup, line 44: I am not a big fan of nesting too many objects, what do you think about just flatting out `client` and `info` on the test struct?

pkg/git/bitbucket/repository_test.go
```diff
@@ -0,0 +1,470 @@
...
+		version string
+	}
+	tests := []struct {
+		name    string
+		in      in
+		want    string
+		wantErr bool
+	}{
+		{
+			name: "Run with success",
```
henriquemoraeszup, line 112: Should we have a server fail case here as well?

pkg/git/bitbucket/repository_test.go
```diff
@@ -0,0 +1,470 @@
...
+			}
+			if !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("LatestTag() got = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
+const (
+	PayloadListAllTags = `{
```
henriquemoraeszup, line 211: Nice 👏

pkg/git/bitbucket/repository_test.go
```diff
@@ -0,0 +1,470 @@
...
+	return m.latestTagUrl
+}
+
+func (m RepoInfoCustomMock) TokenHeader() string {
+	return m.tokenHeader
+}
+
+func (m RepoInfoCustomMock) Token() string {
+	return m.token
+}
```
henriquemoraeszup, line 470: How about we use testify mocks?

pkg/git/bitbucket/bitbucket.go
```diff
@@ -0,0 +1,84 @@
...
+// LatestTagUrl returns the Bitbucket API URL for get latest tag release
+// e.g. https://api.bitbucket.org/2.0/repositories/{{owner}}/{{repo}}/refs/tags?sort=target.date
+func (in DefaultRepoInfo) LatestTagUrl() string {
+	return fmt.Sprintf(LatestTagUrlPattern, in.owner, in.repo)
+}
+
+// TokenHeader returns the Authorization value formatted for Bitbucket API integration
+// e.g. "m6ioSP4o4q6tmiinXHf9KOHxQsjbvShS-zMullcRsxiJtMHIqr2tuHpJZbl-UXpm7E-1meNlAqKzmORTxyoNAnXWZlCFPsvJpQj4evGtafuH4NBBgRrQ_Mc3"
+func (in DefaultRepoInfo) TokenHeader() string {
+	return in.token
```
brunasilvazup, line 79: In github, this is returned with help message, but in the gitlab don't.
You considered this?

brunasilvazup (dismissed):

pkg/git/bitbucket/bitbucket.go
lucasdittrichzup: yep :smile:

pkg/git/bitbucket/repository_test.go
lucasdittrichzup: in use :rocket:

pkg/git/bitbucket/repository_test.go
lucasdittrichzup: :rocket:

pkg/git/bitbucket/repository_test.go
lucasdittrichzup: done :+1:

pkg/git/bitbucket/repository_test.go
lucasdittrichzup: done :+1:

pkg/git/bitbucket/repository_test.go
lucasdittrichzup: done :+1:

pkg/git/bitbucket/bitbucket_test.go
lucasdittrichzup: done :+1:

pkg/git/bitbucket/bitbucket_test.go
lucasdittrichzup: done :+1:

pkg/git/bitbucket/bitbucket.go
lucasdittrichzup: done :+1:

pkg/git/bitbucket/bitbucket.go
lucasdittrichzup: done :+1:

pkg/git/bitbucket/bitbucket.go
lucasdittrichzup: as we talked, thinking about the user experience, it is better to validate that way and he receives a 404 error, than the panic error that occurs when the URL is outside the expected standard :smile:

pkg/git/bitbucket/repository_test.go
lucasdittrichzup: done and thanks for helping :fist_oncoming:

pkg/git/bitbucket/repository.go
lucasdittrichzup: the bitbucket api does not return other 2xx codes (at least the endpoints I use)

pkg/git/bitbucket/bitbucket.go
lucasdittrichzup: since I'm using the RepoInfo interface, I need to return this method, even if it's the same as the token. Otherwise I would have to duplicate a lot of code to meet the bitbucket implementation. However, I believe that we should restructure this later, since the gitlab falls at that point

pkg/git/bitbucket/repository.go
lucasdittrichzup: as I said earlier, I needed to implement TokenHeader to meet the RepoInfo interface. However, I will no longer use it here because I understand that it is not really necessary

pkg/git/bitbucket/bitbucket_test.go
```diff
@@ -0,0 +1,102 @@
...
+	repoInfo := NewRepoInfo("https://bitbucket.org/username/ritchie-formulas/src/master/", "some_token")
+	tagsUrl := repoInfo.TagsUrl()
+
+	assert.Equal(t, want, tagsUrl)
+}
+
+func TestZipUrl(t *testing.T) {
+	const want = "https://bitbucket.org/username/ritchie-formulas/get/1.0.0.zip"
+	repoInfo := NewRepoInfo("https://bitbucket.org/username/ritchie-formulas/src/master/", "some_token")
+	tagsUrl := repoInfo.ZipUrl("1.0.0")
```
brunasilvazup, line 75: Typo `tagsUrl` -> `zipUrl`

pkg/git/bitbucket/repository.go
```diff
@@ -0,0 +1,149 @@
...
+		req.Header.Add(headers.Authorization, fmt.Sprintf("Bearer %s", info.Token()))
+	}
+
+	resp, err := re.client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		defer resp.Body.Close()
```
brunasilvazup, line 64: Should that be out of if?

brunasilvazup, line 64: Should that be out of if?

pkg/git/bitbucket/repository.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	defer res.Body.Close()
+
+	if res.StatusCode != http.StatusOK {
+		b, err := ioutil.ReadAll(res.Body)
+		if err != nil {
+			return git.Tag{}, err
+		}
+		return git.Tag{}, errors.New(string(b))
```
brunasilvazup, line 134: For each method you use a different response, can you unify all error messages on different code than expected?

I liked more:

```go
if resp.StatusCode != http.StatusOK {
		all, _ := ioutil.ReadAll(resp.Body)
		fmt.Println(all)
		return nil, errors.New(resp.Status)
}
```

But i would change `fmt.Println(all)` to next to `errors.New`, something like `(resp.Status + " - " + all)`.

What do you think?

brunasilvazup (commented):
I left some other comments, see what you think.

pkg/git/bitbucket/repository.go
lucasdittrichzup: it's like that, because if I put out the if, in case the status code is 200, the body is closed before downloading the repo

pkg/git/bitbucket/repository.go
lucasdittrichzup: following the tip from @brunasilvazup, I decided to standardize this part of the code in all methods, so this part was removed

brunasilvazup (dismissed):

pkg/git/bitbucket/repository.go
```diff
@@ -0,0 +1,144 @@
...
+	if info.Token() != "" {
+		req.Header.Add(headers.Authorization, fmt.Sprintf("Bearer %s", info.Token()))
+	}
+
+	res, err := re.client.Do(req)
+	if err != nil {
+		return git.Tags{}, err
+	}
+
+	defer res.Body.Close()
```
henriquemoraeszup, line 88: It seems this snippet of code is the same for `Tags`, `Zipball`, and `LatestTag`, how about extracting it into a method `performRequest`.

henriquemoraeszup (commented):
Nice tests, 👏 
Just a couple of comments

internal/mocks/mocks.go
```diff
@@ -382,3 +381,32 @@ func (w *WorkspaceMock) List() (formula.Workspaces, error) {
...
+
+func (ri *RepoInfo) TokenHeader() string {
+	args := ri.Called()
+	return args.String(0)
+}
+
+func (ri *RepoInfo) Token() string {
+	args := ri.Called()
+	return args.String(0)
+}
```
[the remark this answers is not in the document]
lucasdittrichzup: done!

pkg/git/bitbucket/repository.go
lucasdittrichzup: excellent idea! done!

brunasilvazup (approved):

henriquemoraeszup (approved):

PR #824, opened 2021-01-10
Set credential flags

henriquemoraeszup:
### Description

* Adding flag capacity to the command `rit set credential` with provider, fields, and values. Fields and values are a comma separated list of strings. 
* Simplified the user flow by removing the "add new credential from filepath" option. It does not seem to bring any value to provide a filepath for a txt file with a single credential value in it. Also there has not been any reports about user consuming this functionality

### How to verify it
Run `rit set credential --provider=github fields=username,token values=user1,token1`

### Changelog

Added flag capacity to `rit set credential`
Removed add new credential from filepath capacity

Screen Shot 2021-01-10 at 10 53 01


pkg/cmd/set_credential.go
```diff
@@ -147,50 +179,64 @@ func (s setCredentialCmd) prompt() (credential.Detail, error) {
...
-			}
-
-			if !s.FileReadExister.Exists(path) {
-				return credDetail, prompt.NewError("Cannot find any credential file at " + path)
-			}
-
-			byteValue, err := s.FileReadExister.Read(path)
-			if err != nil {
-				return credential.Detail{}, err
-			}
```
henriquemoraeszup, line 166: Intentionally removed the feature to add new credential by file path, since there is no proven point on its usefulness

pkg/cmd/delete_credential_test.go
```diff
@@ -315,31 +315,35 @@ func TestDeleteCredentialFormula(t *testing.T) {
 		inputBoolResult bool
 		inputListError  error
 		fileShouldExist bool
-		args            string
+		args            []string
 		wantErr         string
 	}{
 		{
 			name:            "execute prompt with success",
+			args:            []string{},
```
brunasilvazup, line 323: This arg is really necessary? Or the default of type is equal?

pkg/cmd/delete_credential_test.go
```diff
@@ -315,31 +315,35 @@ func TestDeleteCredentialFormula(t *testing.T) {
...
 		{
 			name:            "execute flag with empty provider fail",
-			args:            "--provider=",
+			args:            []string{"--provider="},
 			wantErr:         "please provide a value for 'provider'",
 			fileShouldExist: true,
 		},
 		{
 			name:            "fail on input list error",
+			args:            []string{},
```
brunasilvazup, line 339: Same question in here.

pkg/cmd/delete_credential_test.go
```diff
@@ -315,31 +315,35 @@ func TestDeleteCredentialFormula(t *testing.T) {
...
 			wantErr:         "some error",
 			inputListError:  errors.New("some error"),
 			fileShouldExist: true,
 		},
 		{
 			name:            "do nothing on input bool refusal",
+			args:            []string{},
```
brunasilvazup, line 346: And here.

brunasilvazup (approved):

pkg/cmd/delete_credential_test.go
henriquemoraeszup: the default is a `nil` value, which breaks the code

pkg/cmd/delete_credential_test.go
henriquemoraeszup: same

pkg/cmd/set_credential.go
```diff
@@ -147,50 +179,62 @@ func (s setCredentialCmd) prompt() (credential.Detail, error) {
...
+func (s *setCredentialCmd) resolveFlags(cmd *cobra.Command) (credential.Detail, error) {
+	provider, err := cmd.Flags().GetString(providerFlagName)
+	if err != nil || provider == "" {
+		return credential.Detail{}, errors.New("please provide a value for 'provider'")
+	}
+
+	fields, err := cmd.Flags().GetStringSlice(fieldsFlagName)
+	if err != nil {
+		return credential.Detail{}, err
+	} else if len(fields) == 0 {
```
lucasdittrichzup, line 212: if it is just an if it will not have the same effect?
```go
if len(fields) == 0  { ... }
```

lucasdittrichzup (approved):

pkg/cmd/set_credential.go
henriquemoraeszup: yup, same thing, it is just a matter of style, I like to join all the if conditions pertaining to the same line of code

PR #826, opened 2021-01-12
Add flags to delete env

brunasilvazup:
### Description

This PR deprecated stdin and add flags to `rit delete env`.

### How to verify it
Clone this PR, build the rit and use `rit delete env --env=prod`

> foto1

### Changelog

Add suport to flag in rit delete env command

### DIFF
--- a/pkg/cmd/delete_env.go
+++ b/pkg/cmd/delete_env.go
@@ -48,7 +48,7 @@ var (
 	}
 )
 
-type inputDeleteEnv struct {
+type teste struct {
 	env string
 }
 
@@ -89,16 +89,16 @@ func NewDeleteEnvCmd(
 
 func (d *deleteEnvCmd) runFormula() CommandRunnerFunc {
 	return func(cmd *cobra.Command, args []string) error {
-		inputParams, err := d.resolveInput(cmd)
+		env, err := d.resolveInput(cmd)
 		if err != nil {
 			return err
 		}
 
-		if inputParams.env == "" {
+		if env == "" {
 			return nil
 		}
 
-		if _, err := d.env.Remove(inputParams.env); err != nil {
+		if _, err := d.env.Remove(env); err != nil {
 			return err
 		}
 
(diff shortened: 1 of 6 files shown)


pkg/cmd/delete_env.go
```diff
@@ -28,11 +30,28 @@ import (
 	"github.com/ZupIT/ritchie-cli/pkg/stdin"
 )
 
+const (
+	envFlagName        = "env"
+	envFlagDescription = "Env name to delete"
+)
+
 var (
```
henriquemoraeszup, line 38: shouldnt these be a const?

pkg/cmd/delete_env.go
```diff
@@ -28,11 +30,28 @@ import (
...
+			kind:        reflect.String,
+			defValue:    "",
+			description: envFlagDescription,
+		},
+	}
 )
 
+type inputDeleteEnv struct {
+	env string
+}
```
henriquemoraeszup, line 53: Since you only need the env name, why not pass around a `string` instead of creating a struct?

pkg/cmd/delete_env.go
```diff
@@ -132,3 +133,57 @@ func (d deleteEnvCmd) runStdin() CommandRunnerFunc {
...
+	return d.resolvePrompt()
+}
+
+func (d *deleteEnvCmd) resolvePrompt() (inputDeleteEnv, error) {
+	envHolder, err := d.env.Find()
+	if err != nil {
+		return inputDeleteEnv{}, err
+	}
+
+	if len(envHolder.All) <= 0 {
```
henriquemoraeszup, line 150: nitpick: there is no possible way a length is < 0. So you can use `==`, also length == 0 translates to a good reader as "empty list"

pkg/cmd/delete_env.go
brunasilvazup: Hmm yes, but the object 'flag' can't be a constant, rules of go.

I can exchange messages for constants, do you think?

pkg/cmd/delete_env.go
brunasilvazup: I did it by default, but looking here I didn't use it anywhere, I'll remove it for simplicity thanks (:

pkg/cmd/delete_env.go
henriquemoraeszup: yeah, not the objects, just the texts

henriquemoraeszup (approved):
🚀

pkg/env/setter_test.go
```diff
@@ -128,7 +132,7 @@ func TestSet(t *testing.T) {
 			in := tt.in
 			out := tt.out
 
-			setter := NewSetter(tmp, in.envFinder, in.file)
+			setter := NewSetter(ritHomeDir, in.envFinder, in.file)
 			got, err := setter.Set(in.env)
 
 			if out.err != nil && out.err.Error() != err.Error() {
```
lucasdittrichzup, line 138: Would it be too complicated to use testfy here?

lucasdittrichzup (approved):

pkg/env/setter_test.go
henriquemoraeszup: agree

brunasilvazup: It is not very complicated if it were just a few changes, but as it is not the focus of PR I preferred to leave it to another specific PR.

PR #833, opened 2021-01-22
Credential for adding Private Repo

matheussn:
### Description

Now the command `rit add repo` has a specific credential, so that the user does not need to regenerate his private tokens each time the command is executed.

### How to verify it
In its first execution, when adding a private repository, the command will request the token, as it already did. However, the command now saves this token in a credential.
After executing the `rit add repo` command adding a new private repository, use the `rit list credential` command, to verify that a new credential has been created in the current environment.

### Changelog

Save and reuse repo credentials for private repos


pkg/cmd/add_repo.go
```diff
@@ -142,7 +143,7 @@ func (ad addRepoCmd) runPrompt() CommandRunnerFunc {
 
 		var token string
 		if isPrivate {
-			token, err = ad.Password("Personal access tokens:")
+			token, err = ad.Resolve("CREDENTIAL_" + provider + "-add-repo_token")
```
matheussn, line 146: I didn't really like this implementation, but I'm afraid to try a more "beautiful" implementation and end up getting more complex than necessary.

I accept suggestions.

henriquemoraeszup:
Hey, we had a recent merge that affects add repo, can you update it?

pkg/cmd/add_repo.go
henriquemoraeszup: I think reusing the resolver is the way to go, if anything changes on the logic, we still have one source of truth.

henriquemoraeszup: On another note, it is convention to use credential constants always capitalized, like `CREDENTIAL_GITHUB_ADD_REPO_TOKEN`

pkg/cmd/add_repo.go
```diff
@@ -153,6 +154,13 @@ func (ad addRepoCmd) runPrompt() CommandRunnerFunc {
 		gitRepoInfo := git.NewRepoInfo(url, token)
 		tags, err := git.Repos.Tags(gitRepoInfo)
 		if err != nil {
+			if strings.Contains(err.Error(), "401") {
+				errorString := fmt.Sprintf("permission error:\nYou must overwrite the current token (%s-add-repo) with command:\n"+
+					"\t rit set credential\n"+
+					"Or move to a new environment with command:\n"+
```
henriquemoraeszup, line 160: Nitpick:
```suggestion
					"Or switch to a new environment with the command:\n"+
```

pkg/credential/settings.go
```diff
@@ -215,13 +215,15 @@ func NewDefaultCredentials() Fields {
...
-		"ansible":    []Field{username, password},
+		AddNew:            []Field{},
+		"github":          []Field{username, email, token},
+		"gitlab":          []Field{username, email, token},
+		"aws":             []Field{accessKey, secretAccessKey},
+		"jenkins":         []Field{username, token},
+		"kubeconfig":      []Field{base64config},
+		"ansible":         []Field{username, password},
+		"github-add-repo": []Field{token},
+		"gitlab-add-repo": []Field{token},
```
henriquemoraeszup, line 226: This is a "reserved" credential, not sure we should expose to the user, what do you think? Can we edit credentials in Ritchie after it has been set?

matheussn: I think it makes sense to keep them "reserved"

brunasilvazup:
> ### Description
> Now the command `rit add repo` has a specific credential, so that the user does not need to regenerate his private tokens each time the command is executed.
> 
> ### How to verify it
> In its first execution, when adding a private repository, the command will request the token, as it already did. However, the command now saves this token in a credential.
> After executing the `rit add repo` command adding a new private repository, use the `rit list credential` command, to verify that a new credential has been created in the current environment.
> 
> ### Changelog
> Save and reuse repo credentials for private repos

Hi @matheussn, could you add gifs with the simulation of your change operation please?

pkg/cmd/add_repo.go
```diff
@@ -235,6 +236,13 @@ func (ar *addRepoCmd) resolvePrompt() (formula.Repo, error) {
 	gitRepoInfo := git.NewRepoInfo(url, token)
 	tags, err := git.Repos.Tags(gitRepoInfo)
 	if err != nil {
+		if strings.Contains(err.Error(), "401") {
+			errorString := fmt.Sprintf("permission error:\nYou must overwrite the current token (%s-add-repo) with command:\n"+
```
kaduartur, line 240: could you extract `errorString` to const and create it using the raw string literals **``**?

pkg/cmd/add_repo.go
```diff
@@ -235,6 +236,13 @@ func (ar *addRepoCmd) resolvePrompt() (formula.Repo, error) {
 	gitRepoInfo := git.NewRepoInfo(url, token)
 	tags, err := git.Repos.Tags(gitRepoInfo)
 	if err != nil {
+		if strings.Contains(err.Error(), "401") {
```
kaduartur, line 239: I think you can create a specific error for 401 Unauthorized and check if err is equal

kaduartur (changes requested):

lucasdittrichzup:
what's up @matheussn? hope you're well! if you have any questions to adjust the PR according to the team's suggestions, just get in touch with us, okay?

GuillaumeFalourd:
Giving continuity to the feature implementation here: https://github.com/ZupIT/ritchie-cli/pull/986

Issue #788, opened 2020-12-15
Tests with Testify in Pkg/Formula/Runner


brunasilvazup:
**What would you like to be added**:
- [x] Config runner // Henrique #836
- [ ] Executor
- [ ] Input resolver
- [ ] Post run
- [x] Pre run builder // Bruna #829

**Why is this needed**:

Issue #776, opened 2020-12-14
Tests with Testify in Pkg/Cmd


brunasilvazup:
**What would you like to be added**:
- [x] Add repo
- [x] Add workspace // Aron #809 
- [ ] Add
- [ ] Autocomplete
- [ ] Build formula
- [ ] Build
- [x] Create formula // Kadu #805 
- [ ] Create
- [x] Delete credential // Henrique #774 
- [x] Delete env // Bruna #822
- [ ] Delete formula
- [x] Delete repo  // Fernando #908 
- [ ] Delete
- [x] Delete workspace // Henrique #832
- [ ] Formula
- [ ] Init
- [ ] List credential
- [ ] List repo
- [ ] List
- [ ] List workspace
- [ ] Metrics
- [ ] Root
- [x] Set credential // Henrique #824
- [x] Set env // Fernando #926 
- [x] Set formula runner // Henrique #836
- [ ] Set priority
- [ ] Set
- [ ] Show env
- [ ] Show formula runner
- [ ] Show
- [x] Tutorial // Henrique #846
- [ ] Update repo
- [ ] Update
- [ ] Upgrade

**Why is this needed**:

PR #836, opened 2021-01-26
Set formula runner flags

henriquemoraeszup:
### Description

* Adding flag capacity to `rit set formula-runner`
* Removed file manager dependency on the config runner
* Refactoring tests on the config runner (contributes to #788)
* Refactored tests on the formula runner command (Contributes to #776)

### How to verify it
Run a flag command with `rit set formula-runner --runner=docker`

### Changelog

Added flag support for rit set formula-runner

### DIFF
--- a/pkg/cmd/set_formula_runner.go
+++ b/pkg/cmd/set_formula_runner.go
@@ -36,10 +36,10 @@ type setFormulaRunnerCmd struct {
 
 var setFormulaRunnerFlags = flags{
 	{
-		name:        providerFlagName,
+		name:        runnerFlagName,
 		kind:        reflect.String,
 		defValue:    "",
-		description: "provider name (i.e.: github)",
+		description: "runner name (local or docker)",
 	},
 }
 
--- a/pkg/cmd/set_formula_runner.go
+++ b/pkg/cmd/set_formula_runner.go
@@ -18,7 +18,9 @@ package cmd
 
 import (
 	"errors"
+	"fmt"
 	"reflect"
+	"strings"
 
 	"github.com/spf13/cobra"
 
@@ -39,7 +41,7 @@ var setFormulaRunnerFlags = flags{
 		name:        runnerFlagName,
 		kind:        reflect.String,
 		defValue:    "",
-		description: "runner name (local or docker)",
+		description: fmt.Sprintf("runner name (%s)", strings.Join(formula.RunnerTypes, "|")),
 	},
 }
 
--- a/pkg/cmd/set_formula_runner.go
+++ b/pkg/cmd/set_formula_runner.go
@@ -17,6 +17,9 @@
 package cmd
 
 import (
+	"errors"
+	"reflect"
+
 	"github.com/spf13/cobra"
 
 	"github.com/ZupIT/ritchie-cli/pkg/formula"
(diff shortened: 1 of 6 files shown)


pkg/cmd/set_formula_runner_test.go
```diff
@@ -18,95 +18,97 @@ package cmd
...
-func Test_setFormulaRunnerCmd(t *testing.T) {
-	type in struct {
-		config formula.ConfigRunner
-		input  prompt.InputList
-	}
+func TestSetFormulaRunnerCmd(t *testing.T) {
+	tmpDir := os.TempDir()
+	ritHome := filepath.Join(tmpDir, "runner")
+	err := os.Mkdir(ritHome, os.ModePerm)
+	assert.NoError(t, err)
```
brunasilvazup, line 38: I think this error checking is unnecessary, in line 91 it's not checked and it is also a small part of the test construction.
What do you think?

brunasilvazup, line 38: I think this error checking is unnecessary, in line 91 it's not checked and it is also a small part of the test construction.
What do you think?

pkg/cmd/set_formula_runner_test.go
```diff
@@ -18,95 +18,97 @@ package cmd
...
-			}
+			err := cmd.Execute()
 
-			if err := setFormulaStdin.Execute(); (err != nil) != tt.wantErr {
-				t.Errorf("set formula runner type stdin command error = %v, wantErr %v", err, tt.wantErr)
+			if err != nil {
+				assert.Equal(t, tt.err, err)
+			} else {
+				assert.Nil(t, tt.err)
+				assert.FileExists(t, runnerFile)
```
brunasilvazup, line 111: Do you can check the value in runnerFile too?

pkg/cmd/set_formula_runner_test.go
```diff
@@ -18,95 +18,97 @@ package cmd
...
+			inputList.On(
+				"List", "Select a default formula run type", mock.Anything, mock.Anything,
+			).Return(tt.runner, tt.listErr)
+
+			cmd := NewSetFormulaRunnerCmd(configManager, inputList)
+			cmd.SetArgs(tt.args)
 
-			setFormulaPrompt.PersistentFlags().Bool("stdin", false, "input by stdin")
-			setFormulaStdin.PersistentFlags().Bool("stdin", true, "input by stdin")
+			cmd.PersistentFlags().Bool("stdin", tt.inputStdin != "", "input by stdin")
```
brunasilvazup, line 100: :1st_place_medal:

pkg/cmd/set_formula_runner_test.go
```diff
@@ -18,95 +18,97 @@ package cmd
...
+			args:       []string{},
 			inputStdin: "{\"runType\": \"invalid\"}\n",
+			err:        ErrInvalidRunType,
 		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			setFormulaPrompt := NewSetFormulaRunnerCmd(tt.in.config, tt.in.input)
-			setFormulaStdin := NewSetFormulaRunnerCmd(tt.in.config, tt.in.input)
+			_ = os.Remove(runnerFile)
```
brunasilvazup, line 91: This is necessary?

pkg/cmd/set_formula_runner_test.go
henriquemoraeszup: Sounds good. Actually, it could break if we actually had this folder already. Even with an error, it is better to ignore it

pkg/cmd/set_formula_runner_test.go
henriquemoraeszup: Yes, since we are creating the file on success runs, we want to ensure it is not there in the first place. Otherwise, a second success could lead to a false positive since the file was already created by a previous test

brunasilvazup (dismissed):

lucasdittrichzup (approved):

pkg/cmd/set_formula_runner.go
```diff
@@ -17,50 +17,56 @@
...
 	config formula.ConfigRunner
 	input  prompt.InputList
 }
 
+var setFormulaRunnerFlags = flags{
+	{
+		name:        runnerFlagName,
+		kind:        reflect.String,
+		defValue:    "",
+		description: "runner name (local or docker)",
```
kaduartur, line 42: Could you use `formula.RunnerTypes` to show the available run types?

kaduartur (changes requested):

kaduartur (approved):
LGTM 🚀

brunasilvazup (approved):
LGTM

lucasdittrichzup (approved):

brunasilvazup:
QA tests on linux and windows are OK.

PR #842, opened 2021-01-28
fix: required multiselect

### LINKED ISSUE #839: Required field in multiselect input
**What happened**:
When entering an input of type *multiselect* without the required field, an error occurs.

**What you expected to happen**:
That the required field is optional when the type is multiselect

**How to reproduce it (as minimally and precisely as possible)**:
1. create a formula with an input of type multselect without the required field

**Anything else we need to know?**:
No.

**Environment**:
- Ritchie version (use `rit --version`): 2.8.0
- Operating System: Linux
- Network plugin / Tool and version (if this is a network-related / tool bug): no
- Others: no

lucasdittrichzup:
### Description
The adoption team found a bug in the multiselect input when the required field is not sent.

To solve this, a validation was added that checks if the field is sent before being used in the code.

Close #839

### How to verify it
1. clone this PR (**`git fetch origin pull / 842 / head: fix/required-multiselect`**)
2. create a formula with **multiselect input** _without_ the **required field**
3. run the formula and verify that the error no longer occurs

### Changelog
Add validation in case the required field is not sent in the multiselect input

### DIFF
--- a/pkg/prompt/multiselect.go
+++ b/pkg/prompt/multiselect.go
@@ -19,6 +19,7 @@ package prompt
 import (
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/ZupIT/ritchie-cli/pkg/formula"
+	fInput "github.com/ZupIT/ritchie-cli/pkg/formula/input"
 )
 
 type SurveyMultiselect struct{}
@@ -40,7 +41,7 @@ func (SurveyMultiselect) Multiselect(input formula.Input) ([]string, error) {
 		},
 	}
 
-	if *input.Required {
+	if fInput.IsRequired(input) {
 		multiQs[0].Validate = survey.Required
 	}
 
--- a/pkg/prompt/multiselect.go
+++ b/pkg/prompt/multiselect.go
@@ -19,7 +19,7 @@ package prompt
 import (
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/ZupIT/ritchie-cli/pkg/formula"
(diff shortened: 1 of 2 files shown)


henriquemoraeszup (dismissed):

brunasilvazup (approved):

pkg/prompt/multiselect.go
```diff
@@ -19,6 +19,7 @@ package prompt
 import (
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/ZupIT/ritchie-cli/pkg/formula"
+	fInput "github.com/ZupIT/ritchie-cli/pkg/formula/input"
```
kaduartur, line 22: I think that you don't need to use an alias in this case.

kaduartur (changes requested):

pkg/prompt/multiselect.go
lucasdittrichzup: I used the alias because the name of the package is input and since I already use that name, there is conflict and the build breaks. Do you have any suggestions on how to resolve this?

henriquemoraeszup: I think alias is fine, its either this or rename your `input` variable

lucasdittrichzup: for now, I decided to rename the variable

brunasilvazup (approved):

kaduartur (approved):

henriquemoraeszup (approved):
This will conflict with #843

PR #843, opened 2021-01-28
Support for list input type

### LINKED ISSUE #783: Tests with Testify in Pkg/Formula/Input/Prompt
- [ ] Prompt
**Why is this needed**:

henriquemoraeszup:
### Description

An excuse to refactor `package prompt` tests. Some of the implementations in this PR
* Support for `list` type in the `config.json` file
* Removed mocks from the test file
* Test refactor of the 900+ lines of code to comply with the new implementation method (closes #783 )
* Removed `fileManager` dependency from `prompt`
* Tests actually check if envs were created
* Removed non dynamic entries from table tests
* Removed table test from `Default` test, since it only had a single test

### How to verify it
Run a formula with the config `"type": "list"`, it should work normally

### Changelog

Support for list input type

### DIFF
--- a/pkg/formula/input/prompt/prompt.go
+++ b/pkg/formula/input/prompt/prompt.go
@@ -135,6 +135,11 @@ func (in InputManager) inputTypeToPrompt(items []string, i formula.Input) (strin
 			return in.loadInputValList(items, i)
 		}
 		return in.textValidator(i)
+	case input.ListType:
+		if items == nil || len(items) == 0 {
+			return "", fmt.Errorf(EmptyItems, i.Name)
+		}
+		return in.loadInputValList(items, i)
 	case input.DynamicType:
 		dl, err := in.dynamicList(i.RequestInfo)
 		if err != nil {
@@ -142,7 +147,7 @@ func (in InputManager) inputTypeToPrompt(items []string, i formula.Input) (strin
 		}
 		return in.List(i.Label, dl, i.Tutorial)
 	case input.Multiselect:
-		if len(items) == 0 {
+		if items == nil || len(items) == 0 {
 			return "", fmt.Errorf(EmptyItems, i.Name)
 		}
 		sl, err := in.Multiselect(i)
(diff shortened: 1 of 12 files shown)


brunasilvazup (dismissed):

pkg/formula/input/prompt/prompt.go
```diff
@@ -196,11 +196,9 @@ func (in InputManager) persistCache(formulaPath, inputVal string, input formula.
 			items = items[0:qtd]
 		}
 		itemsBytes, _ := json.Marshal(items)
-		if err := in.file.Write(cachePath, itemsBytes); err != nil {
-			fmt.Sprintln("Write file error")
-			return
+		if err := ioutil.WriteFile(cachePath, itemsBytes, os.ModePerm); err != nil {
+			fmt.Sprintln("Write cache error")
```
kaduartur, line 200: Could you use `fmt.Println("Write cache error")`?

pkg/formula/input/prompt/prompt.go
```diff
@@ -224,27 +222,23 @@ func (in InputManager) loadInputValList(items []string, input formula.Input) (st
 func (in InputManager) loadItems(input formula.Input, formulaPath string) ([]string, error) {
 	if input.Cache.Active {
```
kaduartur, line 223: Could you simplify this function as suggested below?

```suggestion
	if !input.Cache.Active {
		return input.Items, nil
	}
	 
	cachePath := fmt.Sprintf(CachePattern, formulaPath, strings.ToUpper(input.Name))
	fileBytes, err := ioutil.ReadFile(cachePath)
	if err != nil {
		itemsBytes, err := json.Marshal(input.Items)
		if err != nil {
			return nil, err
		}
	
		if err = ioutil.WriteFile(cachePath, itemsBytes, os.ModePerm); err != nil {
			return nil, err
		}
		
		return input.Items, nil
	}
		
	var items []string
	if err := json.Unmarshal(fileBytes, &items); err != nil {
		return nil, err
	}

	return items, nil
```

kaduartur (changes requested):

kaduartur (approved):

brunasilvazup (approved):

PR #847, opened 2021-02-03
Internationalization Ritchie-cli

kaduartur:
### Description
This pull request implements the internationalization of Ritchie-cli:

- We created a new package called i18n which contains the translation solver by ID and a file called `translations.go` which is automatically generated by the `make generate-translation` command, it is worth remembering that when updating the translation files in `resource/i18n` we should always run this command.

- We centralize the Richie-cli configurations in one place, the file responsible for writing and reading these configs is at `/internal/pkg/config/config.go`. The new config file is created at `˜/.rit/configs.toml`.

- Improved the user experience in the `rit init` command, already using the Ritchie translation part.
Here is a demonstration of the git init command:

rit init

### How to verify it
When executing any Ritchie command it should behave like the latest versions, only the `rit init` command that has had visual modifications.

### Changelog
Internationalization Ritchie-cli


brunasilvazup (dismissed):

henriquemoraeszup (changes requested):
Can you update the readme on how to create translations, which files to edit, which order to add a new term (is it alphabetic? by package?)

internal/pkg/config/config_test.go
```diff
@@ -0,0 +1,144 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			config := NewManager(tt.in.ritHome)
+
+			got := config.Write(tt.in.configs)
+			if tt.want != nil {
+				assert.Equal(t, tt.want.Error(), got.Error())
+			}
+
+			if tt.want == nil {
```
brunasilvazup, line 70: `else` here?

brunasilvazup (approved):

internal/pkg/config/config_test.go
kaduartur: Makes a lot of sense rsrs

brunasilvazup (approved):
🥇

henriquemoraeszup (approved):

PR #848, opened 2021-02-04
refactor: deprecated dynamic input

### LINKED ISSUE #813: Deprecate the dynamic list
We should deprecate the dynamic list in favor to #811 

**Why is this needed**:
We will allow for dynamic inputs with formula outputs so this implementation will no longer be necessary

lucasdittrichzup:
### Description
In the coming months, due to the formula output feature, the type of dynamic input will be depreciated. 

Because of this, this PR adds a warning for users who use this type of input in their formulas, as shown in the GIF below:

<p align="center">
</p>

Close #813 

### How to verify it
1. clone this PR (**`git fetch origin pull / 848 / head: feature/deprecated-dynamic-input`**)
2. create a formula with **dynamic input**
3. run the formula and check that the warning is displayed

### Changelog
Adds depreciation warning of dynamic input


pkg/formula/runner/post_run_test.go
```diff
@@ -81,7 +84,7 @@ func TestPostRun(t *testing.T) {
 			name: "error move new files",
 			in: in{
 				file:   fileManagerMock{mErr: errors.New("error to move new files")},
-				dir:    dirManagerMock{},
+				dir:    nil,
```
brunasilvazup, line 87: I think that passing nil here is not necessary, because I think the default of the error is nil.

pkg/formula/runner/post_run_test.go
```diff
@@ -91,31 +94,33 @@ func TestPostRun(t *testing.T) {
...
-			runner := NewPostRunner(tt.in.file, tt.in.dir)
+			removeDir := &mocks.DirManagerMock{}
+			removeDir.On("Remove", mock.Anything).Return(tt.in.dir)
+			runner := NewPostRunner(tt.in.file, removeDir)
 			got := runner.PostRun(tt.in.setup, tt.in.docker)
 
-			if (tt.want != nil && got == nil) || got != nil && got.Error() != tt.want.Error() {
-				t.Errorf("PostRun(%s) got %v, want %v", tt.name, got, tt.want)
-			}
+			assert.Equal(t, tt.want, got)
```
brunasilvazup, line 122: Beautiful, right :star2:

pkg/formula/runner/post_run_test.go
lucasdittrichzup: you are ritght! I removed this param from all tests :smiley:

brunasilvazup (approved):
LGTM

henriquemoraeszup (approved):
Just to clarify, we cannot consider yet this to be the test refactor. We would still have to remove the `filemanagermock` and test for the file system side effects

brunasilvazup (approved):

PR #866, opened 2021-02-18
Update repo fix

### LINKED ISSUE #851: Some repos from rit update repo do not show all its release
**What happened**:
Intermittently, some repos are not listing their releases when we use the command `rit update repo`. This only happens if you select a specific repo.

**What you expected to happen**:
When running `rit add repo`, the releases should always be listed

**Environment**:
- Ritchie version (use `rit --version`): 2.8.0
- Operating System: Linux & MacOS
- Network plugin / Tool and version (if this is a network-related / tool bug):

image (3)

henriquemoraeszup:
### Description

When a user tries to update a specific repo, the cli interrupts the execution right before listing the release versions. This is a bug that is preventing users to update their repos. This was causing due to the existence of local repos for the user. The array used to select the correct repo was the full array with local repos as well. 

Closes #851 

### How to verify it
Run `rit update repo`, select a **specific** repo and have some local repos configured. It should list the releases and update normally.

### Changelog

Fix rit update repo now allows updates of specific repo

Screen Shot 2021-02-18 at 17 04 14

### DIFF
--- a/pkg/cmd/update_repo_test.go
+++ b/pkg/cmd/update_repo_test.go
@@ -73,7 +73,7 @@ func TestUpdateRepoRun(t *testing.T) {
 			in: in{
 				repo: RepositoryListUpdaterCustomMock{
 					list: func() (formula.Repos, error) {
-						return formula.Repos{*repoTest, *repoTest2}, nil
+						return formula.Repos{*repoTest2, *repoTest}, nil
 					},
 					update: func(name formula.RepoName, version formula.RepoVersion) error {
 						return nil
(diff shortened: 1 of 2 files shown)


GuillaumeFalourd (approved):
🚀

pkg/cmd/update_repo_test.go
```diff
@@ -73,7 +73,7 @@ func TestUpdateRepoRun(t *testing.T) {
 			in: in{
 				repo: RepositoryListUpdaterCustomMock{
 					list: func() (formula.Repos, error) {
-						return formula.Repos{*repoTest, *repoTest2}, nil
+						return formula.Repos{*repoTest2, *repoTest}, nil
```
brunasilvazup, line 76: This change it's necessary?

henriquemoraeszup: I have changed because the problem was on having a local repo before an external one, so with that configuration we were witnessing the "happy path". Inverting the position creates the desired error case

brunasilvazup: Hmm, I understand now, in that case it might be interesting to rename repoTest2 to repoLocal and repoTest to repoExternal.

But this is just a 'seasoning', it's your choice :heavy_check_mark:

brunasilvazup (approved):

PR #872, opened 2021-02-23
Improved error msg on invalid value on list input

fabianofernandeszup:
Signed-off-by: Fabiano Chiareto Fernandes <fabiano.fernandes@zup.com.br>

### Description
Improved error message when an invalid value is passed to an input list via input flag or stdin.

When a value is passed with a coding error or hidden character such as line break, the cli informs that the value passed is not valid, but there is no value that it received, making it difficult to identify the error.

In the example below of a run on windows, the accented character gave an encode problem and this was not possible to identify only with the old error message.

### How to verify it

To check the error, I created two bat formulas with input_list with accented items, and called one formula inside the other, so the encoding error generated an error in the execution of the daughter formula, but it was not possible to identify the error.

**Before correction**
image

**After correction**
image

### Changelog
Improved error message when an invalid value is passed to an input list via input flag or stdin.

### DIFF
--- a/pkg/formula/input/flag/flag.go
+++ b/pkg/formula/input/flag/flag.go
@@ -32,7 +32,7 @@ import (
 )
 
 const (
-	errInvalidInputItemsMsg = "only these input items [%s] are accepted in the %q flag"
+	errInvalidInputItemsMsg = "the value [%v] is not valid, only these input items [%s] are accepted in the %q flag"
 )
 
 type InputManager struct {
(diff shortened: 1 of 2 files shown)


GuillaumeFalourd (approved):
LGTM 👍🏼

brunasilvazup (approved):
:+1:

lucasdittrichzup (approved):

Issue #858, opened 2021-02-12
Add a "contains" operator to the conditional input (config.json file)


GuillaumeFalourd:
**What would you like to be added**:

- Add a "contains" operator to the conditional input on `config.json` file. Today, are supported `==`, `!=`, `<`, `>`, `<=`, and `>=`. As explained in the [documentation](https://docs.ritchiecli.io/tutorials/formulas/how-to-implement-a-formula/config-file).

**Why is this needed**:

- Today, if you want to show new inputs with the conditional type, based on a multiselect type input, you have to implement it through prompt (using split for the multiselect variable) inside the formula implementation.

**Example:**

With the`multiselect variable`=`value1|value2|value3|value4`

This implementation would allow to return a specific list of inputs with the conditional input if a user select `value1` or `value3` inside the multiselect variable.

- [X] Check this box if this issue needs a documentation update

PR #874, opened 2021-02-26
Adding feature containsAny, containsAll, containsOnly, notContainsAny and notContainsAll in conditional options

andressaabreuzup:
Signed-off-by: Finzi <andressa.abreu@zup.com.br>
Issue: https://github.com/ZupIT/ritchie-cli/issues/858

### Description
Adding feature containsAny, containsAll, containsOnly, notContainsAny and notContainsAll in conditional options

### How to verify it
Create a formula with the new conditional options.

Example of a config.json with containsAny:

```json
[
	{
		"name": "sample_multiselect",
		"type": "multiselect",
		"items": [
			"item_1",
			"item_2",
			"item_3",
			"item_4"
		],
		"label": "Choose one or more items: ",
		"required": false,
		"tutorial": "Select one or more items for this field."
	},
	{
		"name": "sample_text",
     	        "type": "text",
		"label": "Type : ",
		"default": "test",
		"condition": {
			"variable": "sample_multiselect",
			"operator": "containsAny",
			"value":    "item_2|item_3"
		}
	}
]
```

### Changelog
Adding feature containsAny, containsAll, containsOnly, notContainsAny and notContainsAll in conditional options


pkg/formula/input/input.go
```diff
@@ -74,9 +119,19 @@ func VerifyConditional(cmd *exec.Cmd, input formula.Input) (bool, error) {
...
+	case "containsAny":
+		return valueContainsAny(value, input.Condition.Value), nil
+	case "containsAll":
+		return valueContainsAll(value, input.Condition.Value), nil
+	case "containsOnly":
+		return valueContainsOnly(value, input.Condition.Value), nil
+	case "notContainsAny":
+		return !valueContainsAny(value, input.Condition.Value), nil
+	case "notContainsAll":
+		return !valueContainsAll(value, input.Condition.Value), nil
```
henriquemoraeszup, line 131: I was thinking this feature could work for `string contains` as well, shouldnt it? For example:
Input `carWash`
Conditional value `car` // should return true

Here seems to be handling only array contains

pkg/formula/input/prompt/prompt_test.go
```diff
@@ -689,6 +689,516 @@ func TestMultiselect(t *testing.T) {
 	}
 }
 
+func TestConditionalInputsWithMultiselect(t *testing.T) {
+	tests := []struct {
+		name             string
+		inputJSON        string
+		multiselectValue []string
+		want             error
```
henriquemoraeszup, line 697: You can remove the `want` if you are not expecting an error in any test

pkg/formula/input/prompt/prompt_test.go
```diff
@@ -689,6 +689,516 @@ func TestMultiselect(t *testing.T) {
...
+							"operator": "containsAny",
+							"value":    "item_2|item_3"
+						}
+					}
+				]`,
+			multiselectValue: []string{"item_1", "item_2", "item_3"},
+			want:             nil,
+		},
+		{
+			name: "error multiselect input test with conditional containsAny",
```
henriquemoraeszup, line 731: I would change the `error` word in the test description, it seems like we are expecting a system error. Maybe something like `fail case`, what do you think? Same applies for the other tests as well

pkg/formula/input/prompt/prompt_test.go
```diff
@@ -689,6 +689,516 @@ func TestMultiselect(t *testing.T) {
...
+				iTextDefault,
+				&mocks.InputBoolMock{},
+				&mocks.InputPasswordMock{},
+				iMultiselect,
+			)
+
+			cmd := &exec.Cmd{}
+			got := inputManager.Inputs(cmd, setup, nil)
+
+			assert.Equal(t, tt.want, got)
```
henriquemoraeszup, line 1197: This is testing only if you do not get an error, but does not test if the cases evaluated correctly. You might want to invoke `VerifyConditional` manually and test if the `bool` signature matches your expected response

pkg/formula/input/prompt/prompt.go
```diff
@@ -78,20 +78,21 @@ func NewInputManager(
 	}
 }
 
-func (in InputManager) Inputs(cmd *exec.Cmd, setup formula.Setup, f *pflag.FlagSet) error {
+func (in InputManager) InputsConditionals(cmd *exec.Cmd, setup formula.Setup, f *pflag.FlagSet) (bool, error) {
```
brunasilvazup, line 81: I noticed that the Boolean return parameter is not used in the function called, so is there any reason for it to be returned?

We already had some conditional inputs and in line 97 they are considered, so I couldn't understand the reason for changing this method, could you explain?

pkg/formula/input/input.go
andressaabreuzup: this feature was designed for the case of multiselect, so the string passed by parameter must match exactly with the strings of the multiselect options, and not with substrings.

pkg/formula/input/prompt/prompt_test.go
andressaabreuzup: I change to want a bool, true in case of succes verify conditional and false in case of fail

pkg/formula/input/prompt/prompt_test.go
andressaabreuzup: done

pkg/formula/input/prompt/prompt_test.go
andressaabreuzup: Calling VerifyConditional manually was generating a high complexity for the test since it is called several times and uses the cache of the previous call to finally test the conditional. I was copying the code of the func Inputs in the test and even so I didn't have all the necessary settings, so I thought about returning the conditionPass in the case of the conditional verification

pkg/formula/input/prompt/prompt_test.go
```diff
@@ -682,7 +682,486 @@ func TestMultiselect(t *testing.T) {
...
 			cmd := &exec.Cmd{}
-			got := inputManager.Inputs(cmd, setup, nil)
+			_, got := inputManager.InputsConditionals(cmd, setup, nil)
+
+			assert.Equal(t, tt.want, got)
+		})
+	}
+}
+
+func TestConditionalInputsWithMultiselect(t *testing.T) {
```
brunasilvazup, line 692: In these tests, an alternative to reduce the lines a bit is to put the inputJson of each type of test in variables, but I don't know if that would decrease the reading, @henriquemoraeszup what do you think?

pkg/formula/input/prompt/prompt_test.go
```diff
@@ -682,7 +682,486 @@ func TestMultiselect(t *testing.T) {
...
+						"condition": {
+							"variable": "sample_multiselect",
+							"operator": "containsOnly",
+							"value":    "item_1|item_3"
+						}
+					}
+				]`,
+			multiselectValue: []string{"item_1", "item_2"},
+			want:             false,
+		},
```
brunasilvazup, line 946: Repeated test?

pkg/formula/input/prompt/prompt_test.go
```diff
@@ -682,7 +682,486 @@ func TestMultiselect(t *testing.T) {
...
+							"operator": "containsOnly",
+							"value":    "item_1|item_2"
+						}
+					}
+				]`,
+			multiselectValue: []string{"item_1", "item_2"},
+			want:             true,
+		},
+		{
+			name: "fail: multiselect input test with conditional containsOnly",
```
brunasilvazup, line 855: This and the test below are valid but have the same name, when the tests runs one will overlap the result of the other.
Can you fix the names?

pkg/formula/input/prompt/prompt_test.go
brunasilvazup, line 946: Repeated test?

pkg/formula/input/prompt/prompt_test.go
andressaabreuzup: each inputJson is different for each test, in this case, I think that the code be inside the test makes it easier to read

pkg/formula/input/prompt/prompt_test.go
```diff
@@ -682,7 +682,486 @@ func TestMultiselect(t *testing.T) {
...
+						"condition": {
+							"variable": "sample_multiselect",
+							"operator": "notContainsAll",
+							"value":    "item_2|item_4"
+						}
+					}
+				]`,
+			multiselectValue: []string{"item_2", "item_4"},
+			want:             false,
+		},
```
brunasilvazup, line 1132: Does this test case fail in the same situation as the top one, looking at the logic, right?

andressaabreuzup: No because the top case has an element (3) in the middle of the multiselect options that the user select that does not belong to the conditional parameters (2 and 4). 

I made two similar tests:
one with two conditional options (2 and 4) and three multiselect (2, 3 and 4) 
one with two conditional options (2 and 4) and two multiselect (2 and 4) 

because I check the size of the two arrays (conditional options and multiselect) in the case of notContainsAll, 
because to contain only and exclusively all the elements, the arrays must at least have the same size

pkg/formula/input/prompt/prompt_test.go
andressaabreuzup: There is no repeated test, the tests are a combination of:

- conditional operator (containsAny, containsAll, containsOnly, notConatinsAny and notContainsAll)
- conditional value separated by "|"
- select options of multiselect select by user separated by "|"

pkg/formula/input/prompt/prompt_test.go
andressaabreuzup: Can I write like:

fail: multiselect input test with conditional containsOnly - fail because should contains only item_1 and item_2 and the values selected on multiselect by the user contains item_1, item_2 and item_3?

Im ifraid of to be repetitive

pkg/formula/input/input.go
henriquemoraeszup: I am a bit confused with this approach, we might need more discussing. We are restraining a conditional to a specific input type. It might not always be clear for the user to interpret this. Should we prefix with `multi`? Like `multiselectContainsAll`?

pkg/formula/input/prompt/prompt_test.go
brunasilvazup: As discussed in Call, I think you can add a 'when selected list has more / less elements'

pkg/formula/input/prompt/prompt_test.go
brunasilvazup: Ok, sorry I think I made a mistake in the analysis
Follow the game :video_game:

pkg/formula/input/prompt/prompt_test.go
brunasilvazup: Like talking on Call, I was referring to two other tests that only have different items, but test the same use case. (:

pkg/formula/input/prompt/prompt_test.go
brunasilvazup: Like talking on Call, I was referring to two other tests that only have different items, but test the same use case. (:

pkg/formula/input/prompt/prompt_test.go
brunasilvazup: As we talked at call, I had referred to the line tests (886 and 917) that test the same use case, but exchange the test items.

pkg/formula/input/input.go
andressaabreuzup: Now supports multiselect and string

pkg/formula/input/prompt/prompt_test.go
andressaabreuzup: I invoked VerifyConditional manually and tested if the bool signature matches with my expected response

pkg/formula/input/prompt/prompt.go
andressaabreuzup: I returned the original signature of the method

pkg/formula/input/prompt/prompt_test.go
andressaabreuzup: done

pkg/formula/input/prompt/prompt_test.go
andressaabreuzup: done

pkg/formula/input/prompt/prompt_test.go
andressaabreuzup: done

pkg/formula/input/input.go
```diff
@@ -42,24 +43,101 @@ func HasRegex(input formula.Input) bool {
...
+		if len(splitValue) != len(splitInput) {
+			return false
+		}
+		for _, v := range splitInput {
+			if !containsArray(splitValue, v) {
+				return false
+			}
+		}
+	} else {
+		if !(strings.EqualFold(strings.ToLower(value), strings.ToLower(input))) {
```
brunasilvazup, line 110: The function EqualFold it's case-insensitivity, then i think that functions `toLower` it's unecessary.

And the parentheses involving the `strings.EqualFold([...])` to use `!` are not necessary too.

Do you can check this cases?

andressaabreuzup: done

pkg/formula/input/input.go
```diff
@@ -51,30 +61,100 @@ func inputConditionVariableExistsOnInputList(variable string, inputList formula.
...
+			if !containsArray(splitValue, v) {
+				return false
+			}
+		}
+	} else {
+		if !strings.EqualFold(value, input) {
+			return false
+		}
+	}
+	return true
```
henriquemoraeszup, line 123: I think you can simplify this line
```suggestion
	} else {
		return strings.EqualFold(value, input)
	}
```

pkg/formula/input/flag/flag.go
```diff
@@ -33,6 +33,7 @@ import (
 
 const (
 	errInvalidInputItemsMsg = "the value [%v] is not valid, only these input items [%s] are accepted in the %q flag"
+	TypeSuffix              = "_type"
```
henriquemoraeszup, line 36: What is this used for?

andressaabreuzup: To save the type of the variable without repeating its name in envVal.
Now envVal would look like this, for example:

RIT_SAMPLE_INPUT=item_1
RIT_SAMPLE_INPUT_TYPE=multiselect

The type was necessary because I need to know the type of variables to use the conditional "contains", because it has two different treatments for string and multiselect

pkg/formula/input/input.go
andressaabreuzup: done

pkg/formula/input/prompt/prompt_test.go
```diff
@@ -753,6 +762,851 @@ func TestMultiselect(t *testing.T) {
...
+				inPath,
+			)
+
+			cmd := &exec.Cmd{}
+
+			got := inputManager.Inputs(cmd, setup, nil)
+			result, _ := input.VerifyConditional(cmd, inputs[1], inputs)
+
+			assert.Equal(t, tt.want.error, got)
+			assert.Equal(t, tt.want.conditionalResult, result)
```
henriquemoraeszup, line 1605: nice tests :clap:

pkg/formula/input/flag/flag.go
henriquemoraeszup: got it, just a convention, one underscore seems to make part of the variable name. For instance if I create a formula with `type` then I would get the env `type_type` which might be semantically confusing. Some frameworks like Django use double underscore to simulate accessing a property, how about we use this convention? For instance `RIT_SAMPLE_INPUT__TYPE` tries to represent `RIT_SAMPLE_INPUT.type`

henriquemoraeszup (dismissed):
Hey, just a comment, but looks good 💯

pkg/formula/input/prompt/prompt_test.go
```diff
@@ -753,6 +762,851 @@ func TestMultiselect(t *testing.T) {
 	}
 }
 
+func TestContainsConditionalInputsMultiselect(t *testing.T) {
+	type TestResults struct {
+		error             error
```
brunasilvazup, line 767: An suggestion: this error is all the time nil, so do you can simplify with remove this variable and with `assert.Nil(got)` in line 1218.

pkg/formula/input/prompt/prompt_test.go
```diff
@@ -753,6 +762,851 @@ func TestMultiselect(t *testing.T) {
...
+
+			assert.Equal(t, tt.want.error, got)
+			assert.Equal(t, tt.want.conditionalResult, result)
+		})
+	}
+}
+
+func TestContainsConditionalInputsString(t *testing.T) {
+	type TestResults struct {
+		error             error
```
brunasilvazup, line 1226: The same suggestion in applicable here.

brunasilvazup (approved):
I left some suggestions, but it's up to you.

Independent, great functionality, congratulations :clap:  (:

pkg/formula/input/input.go
```diff
@@ -42,6 +43,15 @@ func HasRegex(input formula.Input) bool {
 	return len(input.Pattern.Regex) > 0
 }
 
+func containsArray(s []string, str string) bool {
```
kaduartur, line 46: Could you replace this function with `sliceutil.Contains`?

pkg/formula/input/input.go
```diff
@@ -51,30 +61,98 @@ func inputConditionVariableExistsOnInputList(variable string, inputList formula.
 	return false
 }
 
+func containsSubstring(s string, substr string) bool {
+	return strings.Contains(s, substr)
+}
+
+func valueContainsAny(inputType string, value string, input string) bool {
+	splitInput := strings.Split(input, "|")
```
kaduartur, line 69: Could you use the constant `MultiselectSeparator` to avoid `"|"`?

pkg/formula/input/input.go
```diff
@@ -51,30 +61,98 @@ func inputConditionVariableExistsOnInputList(variable string, inputList formula.
...
+	if inputType == MultiselectType {
+		splitValue := strings.Split(value, "|")
+		for _, i := range splitInput {
+			if containsArray(splitValue, i) {
+				return true
+			}
+		}
+	} else {
```
kaduartur, line 77: Do you need this `else` statement?

pkg/formula/input/input.go
```diff
@@ -51,30 +61,98 @@ func inputConditionVariableExistsOnInputList(variable string, inputList formula.
...
+func valueContainsAll(inputType string, value string, input string) bool {
+	splitInput := strings.Split(input, "|")
+	if inputType == MultiselectType {
+		splitValue := strings.Split(value, "|")
+		for _, v := range splitInput {
+			if !containsArray(splitValue, v) {
+				return false
+			}
+		}
+	} else {
```
kaduartur, line 96: Do you need this `else` statement?

pkg/formula/input/input.go
```diff
@@ -51,30 +61,98 @@ func inputConditionVariableExistsOnInputList(variable string, inputList formula.
...
+	return true
+}
+
 func VerifyConditional(cmd *exec.Cmd, input formula.Input, inputList formula.Inputs) (bool, error) {
+
 	if input.Condition.Variable == "" {
 		return true, nil
 	}
 
+	var typeValue string
```
kaduartur, line 130: Could you move these variables to before the "for" instruction at line 138?

pkg/formula/input/input.go
```diff
@@ -91,9 +169,19 @@ func VerifyConditional(cmd *exec.Cmd, input formula.Input, inputList formula.Inp
 		return value < input.Condition.Value, nil
 	case "<=":
 		return value <= input.Condition.Value, nil
+	case "containsAny":
```
kaduartur, line 172: Does it make sense to create constants for these cases?

kaduartur (changes requested):

pkg/formula/input/input.go
andressaabreuzup: resolved in call

pkg/formula/input/input.go
andressaabreuzup: done

kaduartur (approved):
LGTM 🚀

Issue #663, opened 2020-11-09
Detect new repo releases with cache


henriquemoraeszup:
Add caching on new repo releases. What would be adequate? 1 hour? 24? 15 minutes?

GuillaumeFalourd:
Once a day LGTM 🚀

kaduartur:
> Once a day LGTM

I agree, once a day 🔝

PR #878, opened 2021-03-03
Add new fields to enable cache in the repositories

kaduartur:
### Description
To decrease the number of requests to check if a repository has a new version, we are implementing a caching approach. Two new fields have been added to the repo structure which are the `LatestVersion` and `Cache`.

### How to verify it
This PR should not change the behavior of any command in the rit

### Changelog
Add new fields to Repo struct


pkg/formula/repo/adder.go
```diff
@@ -63,7 +63,7 @@ func (ad AddManager) Add(repo formula.Repo) error {
 		return err
 	}
 
-	repo.TreeVersion = "v2"
+	repo.TreeVersion = tree.Version
```
brunasilvazup, line 66: image

Great fix (:

pkg/formula/repo/adder.go
```diff
@@ -116,7 +116,7 @@ func setPriority(repo formula.Repo, repos formula.Repos) formula.Repos {
 		for i := range repos {
 			r := repos[i]
 			if repo.Name == r.Name {
-				repos[i].Priority = repo.Priority
+				repos[i] = repo
```
brunasilvazup, line 119: This change doesn't impact the movePosition function?

kaduartur: No, it works perfectly 💯

brunasilvazup (approved):
:top:

pkg/formula/repo.go
```diff
@@ -18,21 +18,35 @@ package formula
...
 import (
 	"sort"
+	"time"
 
 	"github.com/ZupIT/ritchie-cli/pkg/git"
 )
 
-const RepoCommonsName = RepoName("commons")
+const (
+	RepoCacheTime   = 24 * time.Hour
```
henriquemoraeszup, line 27: How about 1 hour?

kaduartur: In issue(#663), we decided to keep the cache for a day

brunasilvazup:
:white_check_mark:  Tested in QA on Ubuntu, it's OK:
 - updated the existing repos with the latest_version and cache information, updated the cache when updating a repository.

PR #884, opened 2021-03-10
Changing conditional to only return an error when the conditional input variable does not exist in the config.json variable list

andressaabreuzup:
Signed-off-by: Finzi <andressa.abreu@zup.com.br>

### Description
Formulas with nested conditional inputs were returning an unnecessary error.
When there were two flows of nested conditionals, for example, consequently one of them was not satisfied. But an error was returned when the variables of the alternative flow conditionals were tested.  The error said that such variables did not exist, however, they existed. They just weren't being filled by belonging to that other flow that didn't happen.

### How to verify it
Test nested conditional inputs with a alternative flow. (Not a error)
Test conditional inputs that not exists. (A error)

### Changelog
Changing conditional  to only return an error when the conditional input variable does not exist in the config.json variable list

### DIFF
--- a/pkg/formula/input/flag/flag.go
+++ b/pkg/formula/input/flag/flag.go
@@ -50,7 +50,7 @@ func (in InputManager) Inputs(cmd *exec.Cmd, setup formula.Setup, flags *pflag.F
 		var inputVal string
 		var err error
 
-		conditionPass, err := input.VerifyConditional(cmd, i)
+		conditionPass, err := input.VerifyConditional(cmd, i, inputs)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


henriquemoraeszup (approved):
Hey @andressaabreuzup ! Nice description, thank you. Really helped understand the problem. I am good to go but it seems there is a lint problem 😄

andressaabreuzup:
> Hey @andressaabreuzup ! Nice description, thank you. Really helped understand the problem. I am good to go but it seems there is a lint problem 😄

It was a spacing problem, I fixed it, now it's 100% :)

brunasilvazup (approved):
:+1:

PR #897, opened 2021-03-18
 Add config to mount volumes via config.json

maurineimirandazup:
### Description
This change allows informing the directory to be mounted as volumes inside the containers.
### How to verify it
Create an array inside the formula's config.json file as shown.
```
(i.e:
"dockerVolumes": [
         "/home/user-name/folder:/mount/folder"
  ]
```
### Changelog
It allows informing directories to be mounted as volumes in formulas executed via docker.


henriquemoraeszup:
Best PR title ever

pkg/formula/runner/docker/pre_run.go
```diff
@@ -220,3 +227,13 @@ func validateDocker(dockerImg string) error {
 
 	return nil
 }
+
+// validate checks if volumes is not nul
```
lucasdittrichzup, line 231: null

pkg/formula/runner/docker/runner.go
```diff
@@ -93,40 +93,39 @@ func (ru RunManager) Run(def formula.Definition, inputType api.TermInputType, ve
...
 
 func (ru RunManager) runDocker(setup formula.Setup, inputType api.TermInputType, verbose bool, flags *pflag.FlagSet) (*exec.Cmd, error) {
-	volume := fmt.Sprintf("%s:/app", setup.Pwd)
-	homeDirVolume := fmt.Sprintf("%s/.rit:/root/.rit", ru.homeDir)
-	var args []string
+	volumes := []string{
+		fmt.Sprintf("%s:/app", setup.Pwd),
+		fmt.Sprintf("%s/.rit:/root/.rit", ru.homeDir),
+	}
+	if setup.Config.Volumes != nil && len(setup.Config.Volumes) != 0 {
```
lucasdittrichzup, line 100: why is this verification (`len(setup.Config.Volumes) != 0`) necessary?

lucasdittrichzup (approved):

pkg/formula/runner/docker/runner.go
maurineimirandazup: Because if there is an empty value, the docker is unable to mount the volume. Resulting in an error in the formula building.

lucasdittrichzup (approved):

pkg/formula/runner/docker/pre_run.go
maurineimirandazup: Solved.

pkg/formula/runner/docker/pre_run_test.go
```diff
@@ -68,6 +68,36 @@ func TestPreRun(t *testing.T) {
...
+				dir:         dirManager,
+			},
+			out: out{
+				want: formula.Setup{
+					Config: config,
+				},
+				wantErr: false,
+				err:     nil,
+			},
+		},
```
brunasilvazup, line 86: This test is the same of `"docker build success"' right?
What do you think about use in this test an formula with docker volumes defined?

pkg/formula/runner/docker/pre_run_test.go
```diff
@@ -274,6 +304,8 @@ func (fi fileManagerMock) Append(path string, content []byte) error {
 const (
 	configJson = `{
   "dockerImageBuilder": "cimg/go:1.14",
+  "dockerVolumes" : [
+  ],
```
brunasilvazup, line 308: And in your tests do you tested with undefined `dockerVolumes` in formulas?

maurineimirandazup: In this case, the value is not undefined. The array just does not carry any value, so the loop does not iterate and has no volume to mount. This test has been implemented.

pkg/formula/runner/docker/pre_run_test.go
maurineimirandazup: This scenario was tested manually. I haven't found a way to test with valid volumes, because Windows and Linux have different directory structures. Do you have any suggestion?

brunasilvazup: Maybe you can use a directory in tmp dir, what do you think?

brunasilvazup (approved):

pkg/formula/runner/docker/pre_run_test.go
maurineimirandazup: The directory is passed through config.json, it is not possible to use tmp dir to mount the volumes for testing.

PR #900, opened 2021-03-24
feat: capture formula output

lucasdittrichzup:
### Description

This PR adds the **first part** of the **formula output feature**.

Thus, it implements the reading of the standard according to the convention (`::output ...`), the sanitization of the data and the recording of them in a **json file** (`output.json`) inside the bin folder of the formula.

With this, every time the formula is changed and needs to be rebuilt, the file will be rewritten, that is, it will always reflect the most recent changes to the formula.

In addition, since it can transmit sensitive data, any output with the `::output` convention will not appear on the terminal for security.
<br>
> Obs .: due to the necessary changes to meet the conventions, the implementation will be "off" in production until all gaps are resolved in **another PR**

### How to verify it

1. check the unit testing pipeline

### Changelog
Add the first part of the formula output feature.


pkg/formula/runner/local/runner.go
```diff
@@ -126,3 +166,53 @@ func (ru RunManager) setEnvs(cmd *exec.Cmd, pwd string, verbose bool) error {
 
 	return nil
 }
+
+func sanitizeData(data []string) []string {
+	sanitizeData := []string{}
+	for i := range data {
+		output := strings.Split(data[i], " ")[1:]
```
GuillaumeFalourd, line 173: This split for space looks dangerous to me. What if the variable has a space inside it?

GuillaumeFalourd (changes requested):
Good job 👍🏼 

I just made an observation regarding the output space split as it seems to me this could be a problem for output variables with spaces.

The rest looks good to me 🚀

pkg/formula/runner/local/runner.go
```diff
@@ -126,3 +166,53 @@ func (ru RunManager) setEnvs(cmd *exec.Cmd, pwd string, verbose bool) error {
...
+
+	return flattenData
+}
+
+func transformData(data []string) map[string]string {
+	transformData := make(map[string]string)
+	for i := range data {
+		test := strings.Split(data[i], "=")
+		key := test[0]
+		value := test[1]
```
fernandobelettizup, line 198: Hey Lucas, how is it going?

I ran into a bug, but I am not sure if we should fix it in the code itself or by convention.

What happens is that when using a argument that has spaces within it (after splitting the data) there will not be a `key `and `value` within the `test` variable, and calling `test[1]` will break the program.

**How to Verify It**
- Create a formula and edit the default code, so that it contains the code you provided:
```
fmt.Printf("::output arg1=%s arg2=%s argN", arg1, arg2, argN)
```
- When testing the new formula, in this case, add a name with spaces:

image

- When execution reaches `transformData`, `data` will look like

image

- Since `data[1]`, has no `=`, when you split it, `test` will not have `length == 2` and you will get a panic message when trying to assign `value`
```
panic: runtime error: index out of range [1] with length 1
```
**Suggestion**
I thought about it and came with some ideas and think that solving this issue will need some more discussion.
1. By convention, we assume (and alert users) that in order to formula output feature work as intended no spaces are allowed in the input. Leaving it up to the user to check for spaces and, i.e, replacing these with dashes or underscores.
2. We check data before spliting it, i.e, looking for `=` sign and in case it is not there, we assume it is a tail of the previous `value`. At first glance it seems to work, but I do not know if it is scalable.

GuillaumeFalourd: A third option could also be to split using another caracter instead of `space` here:     
```
fmt.Printf("::output arg1=%s arg2=%s argN=%s", arg1, arg2, argN)
```

GuillaumeFalourd: Suggestion: use `--` to split. In that case, the print should look like this:

fmt.Printf("::output --arg1=%s --arg2=%s --argN=%s", arg1, arg2, argN)

As the `--` is already used as input flag, it won't be something new for the user.

henriquemoraeszup:
@lucasdittrichzup my suggestion is to turn off this feature, but merge the code. Then, you can implement Cobra's flag parsing feature from this string from another branch just to not make this one too big

brunasilvazup (changes requested):
I have a question, will this output not work for formula executed via docker or will it be implemented in another pr?

pkg/formula/runner/local/runner.go
```diff
@@ -95,9 +98,46 @@ func (ru RunManager) Run(def formula.Definition, inputType api.TermInputType, ve
...
+
 	if err := cmd.Run(); err != nil {
 		return err
 	}
+	<-done
+
+	if len(output) > 0 {
+		sanitizeOutput := sanitizeData(output)
+		flattenOutput := flattenData(sanitizeOutput)
+		transformOutput := transformData(flattenOutput)
```
henriquemoraeszup, line 135: all these methods are creating sequential loops, I wonder if we could have some file `output.go` to isolate the logic and handle output processing and if it can process a line and write into a file, what do you think?

pkg/formula/runner/local/runner.go
```diff
@@ -126,3 +166,53 @@ func (ru RunManager) setEnvs(cmd *exec.Cmd, pwd string, verbose bool) error {
...
+		test := strings.Split(data[i], "=")
+		key := test[0]
+		value := test[1]
+		transformData[key] = value
+	}
+
+	return transformData
+}
+
+func writeOutput(setup formula.Setup, data map[string]string) error {
```
henriquemoraeszup, line 205: I think the `output` file should handle the writing as well

pkg/formula/runner/local/runner.go
lucasdittrichzup: I agree

pkg/formula/runner/local/runner.go
lucasdittrichzup: Yep

pkg/formula/runner/local/runner.go
lucasdittrichzup: as I have other items to adjust, this first part will be off in production and I will upload the necessary changes to another PR

pkg/formula/runner/local/runner.go
lucasdittrichzup: Thanks for the suggestion guys! @GuillaumeFalourd , the convention we adopted is `... arg1={arg1}, argN={argN}` and will be adjusted in another PR

lucasdittrichzup:
> I have a question, will this output not work for formula executed via docker or will it be implemented in another pr?

That's right @brunasilvazup , the idea is to have a local beta and only then implement it for the docker

brunasilvazup (approved):

fernandobelettizup (approved):

GuillaumeFalourd (approved):

PR #908, opened 2021-04-16
Adding flags, delete confirmation and new tests to delete repo

fernandobelettizup:
### Description

Adding flag support to `rit delete repo` command. Some of the refactoring done as well in this PR:
* Added delete confirmation to `rit delete repo` prompt execution.
* Added tests to support flags and changed tests to work with a tmp directory instead of only mocks.

### How to verify it
* Run `rit delete repo --name=<repo>` and check if has been deleted succesfully.
* Run `rit delete repo` and check if the delete confirmation works.

### Changelog

Added flag support for rit delete repo command


GuillaumeFalourd:
It looks like to use the `Input Flags` feature, you removed the `stdin` implementation for the `rit delete repo` command. 
However, as the `stdin` feature is _only_ deprecated, shouldn't you maintain it as well for compatibility between versions?

pkg/cmd/delete_repo.go
```diff
@@ -17,85 +17,123 @@
...
 
 const (
-	deleteSuccessMsg = "Repository %q was deleted with success"
+	deleteSuccessMsg    = "Repository %q was deleted with success"
+	repoFlagDescription = "Repository name to delete"
 )
 
+var deleteRepoFlags = flags{
+	{
+		name:        nameFlagName,
```
brunasilvazup, line 37: I think it would be nice to create a variable here in this file, it's kind of confusing about the value of this var `nameFlagName`...
What do you think?

pkg/cmd/delete_repo.go
```diff
@@ -17,85 +17,123 @@
...
+	}
 
-		err := stdin.ReadJson(os.Stdin, &repo)
-		if err != nil {
-			return err
-		}
+	if len(repos) == 0 {
+		prompt.Warning("You don't have any repositories")
+		return "", nil
+	}
```
brunasilvazup, line 129: I think this error it's necessary be returned, because in `runFormula` we can't be proceed with this error.

pkg/cmd/delete_repo.go
```diff
@@ -99,3 +109,53 @@ func (dr deleteRepoCmd) runStdin() CommandRunnerFunc {
...
+	return repo, nil
+}
+
+func (dr *deleteRepoCmd) resolveFlags(cmd *cobra.Command) (string, error) {
+	name, err := cmd.Flags().GetString(nameFlag)
+	if err != nil {
+		return "", err
+	} else if name == "" {
+		return "", errors.New("please provide a value for 'name'")
```
brunasilvazup, line 157: Here you can use:
```go
errors.New("please provide a value for '" + nameFlag  "'")
```
to reflect the name of the flag and prevent errors in case of changes or `fmt.Errorf(`[...]

pkg/cmd/delete_repo_test.go
```diff
@@ -88,30 +128,63 @@ func TestNewDeleteRepo(t *testing.T) {
...
-				repoManagerMock,
+				inputBoolMock,
+				repoDeleter,
 			)
-			// TODO: remove stdin flag after  deprecation
 			cmd.PersistentFlags().Bool("stdin", false, "input by stdin")
 			cmd.SetArgs(tt.in.args)
 
 			got := cmd.Execute()
+			if got != nil || tt.name == "do not accept delete selected repo" {
```
brunasilvazup, line 183: Do you can use `if tt.want == nil`, so you don't need use two conditions in if, what do you think?

pkg/formula/repo/deleter.go
```diff
@@ -53,6 +55,10 @@ func (dm DeleteManager) Delete(repoName formula.RepoName) error {
 
 func (dm DeleteManager) deleteRepoDir(repoName formula.RepoName) error {
 	repoPath := filepath.Join(dm.ritHome, reposDirName, repoName.String())
+	if _, err := os.Stat(repoPath); os.IsNotExist(err) {
+		return errors.New("no repository with this name")
+	}
+
```
brunasilvazup, line 61: What do you think about adding tests on the `deleter_test.go` on this rule?

pkg/cmd/delete_repo_test.go
```diff
@@ -88,30 +128,63 @@ func TestNewDeleteRepo(t *testing.T) {
...
 				inputListMock,
-				repoManagerMock,
+				inputBoolMock,
+				repoDeleter,
 			)
-			// TODO: remove stdin flag after  deprecation
 			cmd.PersistentFlags().Bool("stdin", false, "input by stdin")
 			cmd.SetArgs(tt.in.args)
 
 			got := cmd.Execute()
```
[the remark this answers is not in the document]
fernandobelettizup: I am not sure if this is a good option.

Right now I am calling the deleter, where I check for the existence of the repo by looking for its path.

In order to achieve this check before command execution, I would have to list all repo names and check if name given on input flag is a part of it.
I am afraid that checking for existence at this point is not scalable.

pkg/cmd/delete_repo_test.go
fernandobelettizup: I am afraid I can not use suggested solution.

For instance, if I use `if tt.want == nil`, when I am testing that the user did not accept to delete selected repo, no errors are expected and deletion should no happen, resulting in the repo directory still existing.
But also all success cases, in which repo directory should no exist, would be filtered by the same condition.
That means that for the same condition 2 results are expected and assertion would not work.

Likewise, the opposite `if tt.want != nil`, it still not works, because in given case there is no expected error.

pkg/cmd/delete_repo_test.go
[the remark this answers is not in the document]
brunasilvazup: About verification, I had thought of a simpler way. I thought about using `assert.DirExists(t, repoPath)` for that, what do you think?

pkg/cmd/delete_repo_test.go
brunasilvazup: I understand now, how about if we use a variable in the test input struct for this:
`mustDelete`
What do you think?

pkg/cmd/delete_repo_test.go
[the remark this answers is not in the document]
brunasilvazup: I saw that there is a test case that tests precisely that there is no repo.
That would make things more complicated, I think we can disregard this suggestion. (:
#FollowTheGame

pkg/cmd/delete_repo_test.go
```diff
@@ -87,31 +117,83 @@ func TestNewDeleteRepo(t *testing.T) {
...
+				args: []string{},
+				repoList: formula.Repos{
+					{
+						Name:     "repoName",
+						Priority: 0,
+					},
+				},
+				inputListString:       "repoName",
+				inputBool:             false,
+				existingRepoIsDeleted: false,
```
henriquemoraeszup, line 138: golang defaults these values to `false` if you do not fill them, just a fact

functional/stdin/stdin_feature.json
```diff
@@ -63,22 +63,6 @@
...
-                "action": "echo"
-            },
-            {
-                "key": "",
-                "value": "delete repo --stdin",
-                "action": "main"
-            }
-        ],
-        "result": "\"Leonidas\" was deleted with success"
-    },
```
brunasilvazup, line 81: This can be returned too, right?

functional/stdin/stdin_integration_test.go
```diff
@@ -56,6 +56,5 @@ var _ = Describe("RitStdin", func() {
 		Entry(scenariosStdin[8].Entry, scenariosStdin[8]),
 		Entry(scenariosStdin[9].Entry, scenariosStdin[9]),
 		Entry(scenariosStdin[10].Entry, scenariosStdin[10]),
-		Entry(scenariosStdin[11].Entry, scenariosStdin[11]),
```
brunasilvazup, line 59: Same comment to here.

brunasilvazup (dismissed):

henriquemoraeszup (approved):

brunasilvazup (approved):
👍🏾

GuillaumeFalourd (approved):
🚀

pkg/cmd/delete_repo.go
```diff
@@ -28,55 +30,63 @@ import (
...
 		Args:      cobra.OnlyValidArgs,
 	}
+
+	addReservedFlags(cmd.Flags(), deleteRepoFlags)
+
 	return cmd
 }
 
-func (dr deleteRepoCmd) runFunc() CommandRunnerFunc {
+func (dr deleteRepoCmd) runFormula() CommandRunnerFunc {
```
kaduartur, line 76: Could you rename the function to `runCmd`?

kaduartur (approved):

Issue #857, opened 2021-02-11
Add command 'rit list formulas'


GuillaumeFalourd:
**What would you like to be added**:

- A core command that allows to list available commands / formulas from a specific repository.

**Why is this needed**:

- When you work with various repositories, it is sometimes difficult to remember in which repository formulas are located. It could be useful to have this feature instead of looking for the formula manually.

**Suggestion**:

When executing the command, the user would have to inform one input inside a list with repository names.

```bash
rit list formulas
all
> commons
demo
formulas-repo
```

Then it would return a list of all the repository formulas with the `help.json` short description:

```
rit aws create bucket: Formula that creates an aws bucket.
rit kafka create topic: Formula that creates a topic on kafka.
...
```

- [X] Check this box if this issue needs a documentation update

PR #913, opened 2021-04-20
Feature/add formula rit list formulas

fernandobelettizup:
### Description

Added command `rit list formula` (#857)
* Command available for prompt and input flags execution.
* Unit tests created for both cases.
* Formulas are listed based on selected repository or all repositories.
### How to verify it
* Run `rit list formula` and follow prompt instructions.
* Run `rit list formula --name=<repo_name>` to check for input flag compatibility.
* Run `rit lit formula --name=ALL` to list formulas from all repositories when using input flags.

Prompt execution
image

Results for selected repository
image
### Changelog

Added command `rit list formula`


pkg/cmd/list.go
```diff
@@ -21,17 +21,17 @@ import "github.com/spf13/cobra"
...
 // NewListCmd create a new list instance.
 func NewListCmd() *cobra.Command {
 	return &cobra.Command{
 		Use:       "list SUBCOMMAND",
-		Short:     "List repositories, credentials, or workspaces",
+		Short:     "List repositories, credentials, workspaces, formulas",
 		Long:      descListLong,
-		Example:   "rit list repo, rit list credential, rit list workspace",
-		ValidArgs: []string{"credential", "repo"},
+		Example:   "rit list repo, rit list credential, rit list workspace, rit list fomrula",
```
GuillaumeFalourd, line 33: * `formula` (you spelled it wrong)

internal/mocks/mocks.go
```diff
@@ -363,6 +363,11 @@ func (t *TreeManager) MergedTree(core bool) formula.Tree {
 	return args.Get(0).(formula.Tree)
 }
 
+func (t *TreeManager) TreeByRepo(formula.RepoName) (formula.Tree, error) {
+	args := t.Called()
```
brunasilvazup, line 367: Do you can add the parameter here?

pkg/cmd/list_formula.go
```diff
@@ -0,0 +1,234 @@
...
+}
+
+func (lr *listFormulaCmd) resolvePrompt() (formula.Repos, error) {
+	repos, err := lr.RepositoryLister.List()
+	if err != nil {
+		return formula.Repos{}, err
+	}
+
+	if len(repos) == 0 {
+		prompt.Warning("You don't have any repositories")
```
lucasdittrichzup, line 129: what do you think about making this sentence a const?

pkg/cmd/list_formula.go
```diff
@@ -0,0 +1,234 @@
...
+		Example: "rit list formula",
+		RunE:    lf.runFormula(),
+	}
+
+	addReservedFlags(cmd.Flags(), listFormulaFlags)
+
+	return cmd
+}
+
+func (lr listFormulaCmd) runFormula() CommandRunnerFunc {
```
brunasilvazup, line 90: Here you could use a pointer `lr *listFormulaCmd`, what you think?

pkg/cmd/list_formula.go
```diff
@@ -0,0 +1,234 @@
...
+	var repoToListFormulas []formula.Repo
+	if repoName == listOptionAll {
+		repoToListFormulas = repos
+	} else {
+		for i := range repos {
+			if repos[i].Name == formula.RepoName(repoName) {
+				repoToListFormulas = append(repoToListFormulas, repos[i])
+				break
+			}
+		}
```
brunasilvazup, line 153: The Repos type has a `Get` method, did you evaluate the use?

pkg/cmd/list_formula.go
```diff
@@ -0,0 +1,234 @@
...
+
+	return len(table.Rows) - 1, nil
+}
+
+func (lr listFormulaCmd) formulasByRepo(repoName formula.RepoName) ([]formulaDefinition, error) {
+	tree, err := lr.treeManager.TreeByRepo(repoName)
+	if err != nil {
+		return []formulaDefinition{}, err
+	} else if tree.Commands == nil {
+		return []formulaDefinition{}, errors.New("no repository with this name")
```
brunasilvazup, line 205: This error it's related with formulas right? So, what do you think to change the error to `errors.New("no formulas in this repository")` or `errors.New("not found formulas in this repository")`

pkg/cmd/list_formula.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	addReservedFlags(cmd.Flags(), listFormulaFlags)
+
+	return cmd
+}
+
+func (lr listFormulaCmd) runFormula() CommandRunnerFunc {
+	return func(cmd *cobra.Command, args []string) error {
+		repos, err := lr.resolveInput(cmd)
```
brunasilvazup, line 92: I noticed that only the names of the repositories are used, how about changing the structure of Repos to [] String?

brunasilvazup, line 92: I noticed that only the names of the repositories are used, how about changing the structure of `Repos` to `[]String`?
And you can extract ALL resolver here too.

pkg/cmd/list_formula.go
fernandobelettizup: Hey Bruna,

I evaluated using it, and it works similarly to this for loop, but as it returns a `formula.Repo`.
In order to make it work, I would have to assign it to a new variable and then append to my slice.
It is pretty much the same, but also I would have to check for an extra error that was already covered when using `InputList`.

I am not sure, but it seems like it is not as fast as just checking for it straight forward.
What do you think about it?

Here is the way I coded in order to use `Get` method
```
repo, err := repos.Get(repoName)
if err != nil {
   return repoToListFormulas, err
}
repoToListFormulas = append(repoToListFormulas, repo)
```

pkg/cmd/list_formula.go
fernandobelettizup: I am afraid in this case not.
`tree.Command` should only be empty if repo does not exist, because `tree.json` could not be found.
That being because when a repo does not have formulas, it cannot be added using `rit add repo`.

But your comment made me realize that maybe the repo exists and `tree.json` could be corrupted, folders might have been deleted, and repo still exists but no formulas could be retrieved.
I will change logic the check for both cases.

pkg/cmd/list_formula.go
fernandobelettizup: The main reason why I am using `Repos` are because `RepositoryLister` return `Repos`.

For prompt execution, changing everything to `String` looks good. Because `repoNames` has already been extracted aiming to fill `InputList` and no further calculation is needed.

But when it comes to flags execution, if flag input is `"ALL"`, I will have to list repos (as it is right now) and also loop into repos to retrieve all repo names before sending this information to `printFormulas`.

That meaning that if I choose to work with `String` instead, I would have an extra loop execution for this specific case. And by using `Repos` this would not happen.

Do you agree with my analysis?

pkg/cmd/list_formula_test.go
```diff
@@ -0,0 +1,168 @@
...
+		},
+		{
+			name: "success prompt option all",
+			in: in{
+				args:            []string{},
+				repoList:        repos,
+				inputListString: "ALL",
+			},
+		},
+		{
```
brunasilvazup, line 86: You can add a test with the 'ALL' option via flags too. What about?

pkg/cmd/list_formula_test.go
```diff
@@ -0,0 +1,168 @@
...
+			name: "success flag",
+			in: in{
+				args:     []string{"--name=repoName"},
+				repoList: repos,
+			},
+		},
```
brunasilvazup, line 92: You can also add a test with the 'ALL' option using flags. What about?

pkg/cmd/list.go
```diff
@@ -21,17 +21,17 @@ import "github.com/spf13/cobra"
...
 func NewListCmd() *cobra.Command {
 	return &cobra.Command{
 		Use:       "list SUBCOMMAND",
-		Short:     "List repositories, credentials, or workspaces",
+		Short:     "List repositories, credentials, workspaces, formulas",
 		Long:      descListLong,
-		Example:   "rit list repo, rit list credential, rit list workspace",
-		ValidArgs: []string{"credential", "repo"},
+		Example:   "rit list repo, rit list credential, rit list workspace, rit list formula",
+		ValidArgs: []string{"credential", "repo", "workspace", "formula"},
```
brunasilvazup, line 34: Hello @fernandobelettizup, can you add the new formula in `Commands`- `api.go` so that the autocomplete works normally?

pkg/cmd/list_formula.go
brunasilvazup: I think I agree with your logic.
#FollowTheGame 👍🏾

pkg/cmd/mocks_test.go
```diff
@@ -381,6 +381,10 @@ func (t treeMock) MergedTree(bool) formula.Tree {
 	return t.tree
 }
 
+func (t treeMock) TreeByRepo(formula.RepoName) (formula.Tree, error) {
+	return t.tree, t.error
+}
+
```
brunasilvazup, line 387: Usually we remove functions from here, this function it's really necessary?

I not found were that is used.

pkg/cmd/list_formula.go
brunasilvazup: I think it does make sense

pkg/cmd/mocks_test.go
fernandobelettizup: If we do not not implement this function on `mocks_test.go`, tests from `add_repo_test` and `formula_test` breaks.
That happens because used `treeMock` implements type `formula.Tree` that needs to have `TreeByRepo` method.

pkg/cmd/list_formula.go
```diff
@@ -0,0 +1,241 @@
...
+		return formula.Repos{}, err
+	}
+
+	if len(repos) == 0 {
+		prompt.Warning(noRepoFoundMsg)
+		return formula.Repos{}, nil
+	}
+
+	reposNames := make([]string, 0, len(repos)+1)
+	reposNames = append(reposNames, listOptionAll)
```
brunasilvazup, line 139: You can simplify here with `reposNames := []string{listOptionAll}`

pkg/cmd/mocks_test.go
brunasilvazup: Oh it's true, right thanks for explanation (:

brunasilvazup (approved):
Good job, everything is fine with me 👍🏾

pkg/cmd/list_formula.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+	return repoToListFormulas, nil
+}
+
+func (lr *listFormulaCmd) resolveFlags(cmd *cobra.Command) (formula.Repos, error) {
+	name, err := cmd.Flags().GetString(nameFlagName)
+	if err != nil {
+		return formula.Repos{}, err
+	} else if name == "" {
+		return formula.Repos{}, errors.New("please provide a value for 'name'")
```
henriquemoraeszup, line 167: we have a method `missingFlagText` for standarization of an empty flag message

pkg/cmd/list_formula.go
```diff
@@ -0,0 +1,240 @@
...
+)
+
+const (
+	listOptionAll       = "ALL"
+	rootString          = "root"
+	rootCommand         = "rit"
+	commandSeparator    = "_"
+	totalFormulasMsg    = "There are %v formulas"
+	totalOneFormulaMsg  = "There is 1 formula"
+	repoFlagDescription = "Repository name to list formulas"
```
henriquemoraeszup, line 41: It is not expected that the user implicitly knows the use of the name `ALL`, you could add this note on the description

pkg/cmd/list_formula.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+func (lr listFormulaCmd) printFormulas(repos formula.Repos) (formulaCount int, err error) {
+	table := uitable.New()
+	table.AddRow("COMMAND", "DESCRIPTION")
+	allFormulas := make([]formulaDefinition, 0)
+	for _, r := range repos {
+		repoFormulas, err := lr.formulasByRepo(r.Name)
+		if err != nil {
+			return 0, err
```
henriquemoraeszup, line 188: If I have a single wrong name, I am aborting the whole process, should you ignore the failure or keep this approach? Just to think about it

pkg/cmd/list_formula.go
```diff
@@ -0,0 +1,240 @@
...
+
+	var repoFormulas []formulaDefinition
+	for _, cmd := range tree.Commands {
+		if cmd.Formula {
+			replacer := strings.NewReplacer(rootString, rootCommand, commandSeparator, " ")
+			parentFormula := replacer.Replace(cmd.Parent)
+
+			var fd formulaDefinition
+			fd.Cmd = strings.Join([]string{parentFormula, cmd.Usage}, " ")
+			fd.Desc = cmd.Help
```
henriquemoraeszup, line 222: I think you could use a constructor. It is a more Gopher way
```
fd := formulaDefinition {
  Cmd: value,
  Desc: value2,
}
```

pkg/cmd/list_formula_test.go
```diff
@@ -0,0 +1,175 @@
...
+			cmd := NewListFormulaCmd(
+				repoManagerMock,
+				inputListMock,
+				treeManager,
+				finderTutorial,
+			)
+			cmd.SetArgs(tt.in.args)
+
+			got := cmd.Execute()
+			assert.Equal(t, tt.want, got)
```
henriquemoraeszup, line 172: Since this command is about displaying a table, what do you think about checking for certain texts on `stdout`? `builder_test` has an example

henriquemoraeszup (changes requested):
Nice tests 🚀 just some points for debate

pkg/cmd/list_formula.go
fernandobelettizup: Let us take a look at the prompt execution:
- User selects repo from repo input list, and we try to print formulas.
- Given that the repo name came from the input list, there should not be any wrong names, since input list only shows existing repos.

The only way we could face a wrong name error is when user tries to list formulas using input flags, in that case we kill the process before printing any rows of the table.

Although, if we are trying to print multiple repo formulas by selecting `ALL` option, it may happen that, when trying to read formulas an error occurs.
In this case what could be better?
Should we save faulty repo name, print all formulas and also a warning before the `[TUTORIAL]` section including a message similar to: 
```
Formulas from repoName could not be retrieved.
```
Or since we had an error when trying to retrieve all formulas we abort the whole process?

pkg/cmd/list_formula.go
fernandobelettizup: I have thought of this message:
```
repoFlagDescription = "Repository name to list formulas, use 'ALL' to list formulas from all repositories."
```
When running `rit list formula --help` this would be the output:
image

Is it ok? Or would it be better to add this note at the description (first line of output)?

henriquemoraeszup: that looks good enough

pkg/cmd/list_formula.go
henriquemoraeszup: if you abort the whole process from a missing repo in ALL, you are subjecting your user to learn how to fix Ritchie's configurations and preventing her from getting the output of the command. Adding a warning would allow the user to opt into correcting the problem without breaking its experience

kaduartur (changes requested):

pkg/cmd/list_formula.go
```diff
@@ -0,0 +1,254 @@
...
+		Example: "rit list formula",
+		RunE:    lf.runFormula(),
+	}
+
+	addReservedFlags(cmd.Flags(), listFormulaFlags)
+
+	return cmd
+}
+
+func (lr *listFormulaCmd) runFormula() CommandRunnerFunc {
```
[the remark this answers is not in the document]
fernandobelettizup: It seems to be a standard using `runFormula` on all files that use input flags.

pkg/cmd/list_formula.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+
+	repoName, err := lr.InputList.List("Repository:", reposNames)
+	if err != nil {
+		return formula.Repos{}, err
+	}
+
+	var repoToListFormulas []formula.Repo
+	if repoName == listOptionAll {
+		repoToListFormulas = repos
```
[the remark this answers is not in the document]
fernandobelettizup: I can return `repos`, but still need else statement just in case list option is not `ALL`, in this situation I would return `repos[i]`, instead of appending to the list.
What do you think about it?

pkg/cmd/list_formula.go
```diff
@@ -0,0 +1,254 @@
...
+	table.AddRow("COMMAND", "DESCRIPTION")
+	allFormulas := make([]formulaDefinition, 0)
+	failedRepos := make([]string, 0)
+	for _, r := range repos {
+		repoFormulas, err := lr.formulasByRepo(r.Name)
+		if err != nil {
+			if len(repos) > 1 {
+				failedRepos = append(failedRepos, r.Name.String())
+			} else {
+				return 0, err
```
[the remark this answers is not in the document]
fernandobelettizup: For both `ALL` and single repo execution, error was converted into a warning message.
For instance, when running for `ALL`, there can happen two warnings, one for unknown error and one for no formulas in repo.
In any cases these warnings prevent process abortion.
image

When running for only one repo, tutorial is also changed:
image

What do you think about these changes?

kaduartur (dismissed):

pkg/cmd/list_formula.go
```diff
@@ -0,0 +1,275 @@
...
+	emptyRepos := make([]string, 0)
+	for _, r := range repos {
+		repoFormulas, err := lr.formulasByRepo(r.Name)
+		if err != nil {
+			if len(repos) == 1 {
+				if err != errEmptyTree {
+					return 0, err
+				}
+				prompt.Warning(fmt.Sprintf(emptyRepoMsg, r.Name.String()))
+			}
```
henriquemoraeszup, line 200: Just curious, would we be able to remove all these conditionals and still get the same behavior? I mean, for a single repo just issue the warning. I know we would execute operations on a single item array like `Sort` but the processing power is negligible, and removing this code would reduce cognitive complexity

pkg/cmd/list_formula_test.go
```diff
@@ -0,0 +1,225 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			inputListMock := new(mocks.InputListMock)
+			inputListMock.On("List", mock.Anything, mock.Anything, mock.Anything).Return(tt.in.inputListString, tt.in.inputListErr)
+
+			for i, r := range tt.in.repoList {
+				repoName := r.Name.String()
+				repoPath := filepath.Join(reposPath, repoName)
+				_ = os.MkdirAll(repoPath, os.ModePerm)
+				_ = streams.Unzip("../../testdata/tree.zip", repoPath)
```
henriquemoraeszup, line 173: Should we join with path or is this ok? Wouldnt this break on windows?

henriquemoraeszup (dismissed):
Approving, but you should check the comments

pkg/cmd/list_formula_test.go
fernandobelettizup: It works just fine on Windows too.

pkg/cmd/list_formula.go
fernandobelettizup: I do not see any problems just issuing the warning for a single repo.
In this case is it ok if I check for `allFormulas` length before running any table operation by moving all table related code to a conditional block?

```
if len(allFormulas) != 0 {
	table := uitable.New()
	table.AddRow("COMMAND", "DESCRIPTION")
	sort.Sort(ByCmd(allFormulas))
	for _, fm := range allFormulas {
		table.AddRow(fm.Cmd, fm.Desc)
	}
	raw := table.Bytes()
	raw = append(raw, []byte("\n")...)
	fmt.Println(string(raw))
}
```

kaduartur (dismissed):

kaduartur (approved):

PR #922, opened 2021-05-12
Add `rit update workspace` core command

GuillaumeFalourd:
### Description

- Add the `rit update workspace` core command to allow users to update the workspace `tree.json` file and access new formulas (or formulas' updates) after updating the workspace using `git` commands.

### How to verify it

- Install Ritchie CLI.
- Use this branch locally.
- Overwrite your current binary following [this guide](https://github.com/ZupIT/ritchie-cli/blob/master/DEVELOPER_GUIDE.md#overwriting-your-current-binary).
- Execute the `rit update workspace` command

### Demo

rit update workspace

https://user-images.githubusercontent.com/22433243/118698408-bd3a2e80-b7e6-11eb-93c3-c25806a1ccae.mp4

### Changelog

- Add the `rit update workspace` core command.

### Tasks

- [X] Implementation
- [X] Unit Tests
- [X] Documentation : https://github.com/ZupIT/docs-ritchie/pull/82


pkg/formula/workspace/workspace.go
```diff
@@ -126,6 +131,36 @@ func (m Manager) Delete(workspace formula.Workspace) error {
...
+		return err
+	}
+
+	if _, exists := workspaces[workspace.Name]; !exists {
+		return ErrInvalidWorkspace
+	}
+
+	workspaceLocalName := repoutil.LocalName(workspace.Name)
+	workflowPath := filepath.Join(m.ritchieHome, formula.ReposDir, workspaceLocalName.String())
+	treeData, err := m.tree.Generate(workflowPath)
```
henriquemoraeszup, line 146: If you get the correct workspace from `workspaces`, doesnt it already have a property `Dir`? I might be able to use it

henriquemoraeszup: I think you are thinking too much about workflows with the variable naming 😝

pkg/cmd/update_workspace.go
```diff
@@ -0,0 +1,143 @@
...
+		items = append(items, kv)
+	}
+
+	selected, err := u.inList.List("Select the workspace to update: ", items)
+	if err != nil {
+		return formula.Workspace{}, err
+	}
+
+	split := strings.Split(selected, " (")
+	workspaceName := split[0]
```
henriquemoraeszup, line 114: do we have a safer way of doing this? Maybe by the index on the list or something

pkg/formula/workspace/workspace.go
```diff
@@ -126,6 +131,36 @@ func (m Manager) Delete(workspace formula.Workspace) error {
...
+	if err != nil {
+		return err
+	}
+
+	if _, exists := workspaces[workspace.Name]; !exists {
+		return ErrInvalidWorkspace
+	}
+
+	workspaceLocalName := repoutil.LocalName(workspace.Name)
+	workflowPath := filepath.Join(m.ritchieHome, formula.ReposDir, workspaceLocalName.String())
```
henriquemoraeszup, line 145: I dont think you would need all these checks because your resolvers on `update_workspace` already ensure you have a real one in hand. Just use the one you get as a parameter

henriquemoraeszup (commented):
Nice, looks like it is on the right place

pkg/cmd/update_workspace.go
GuillaumeFalourd: The List method returns an element like this: `Default (/Users/username/ritchie-formulas-local)`
Therefore this is the easiest way to get the workspace's name based on the user choice (as we don't know how he wrote the workspace's name when he added it).

We could eventually refactor the List method, but it would be necessary to change other core commands implementations. I suggest to do it as another improvement feature PR to not mix things.

Do you agree?

pkg/formula/workspace/workspace.go
GuillaumeFalourd: This is the workspacePath to the `.rit/repos` folder (not the path to the local repo). I will rename the variable to make it clearer 👍🏼

pkg/cmd/update_workspace.go
fernandobelettizup: I have thought of some workarounds that may be ok.

For instance, let us think that we have a workspace named `Fernando (Test)`. With current implementation code will break, because split will give us 3 slices: `Fernando`, `Test)` and `path`.

Hoping that the user does not use any spaces on directories, we can find `workspaceName` by joining the recently split slice, but ignoring the last element, that should be the path.

Another approach is that instead of looking for the `workspaceName` we always look for the `workspacePath` by getting the last element of `split` (and cleaning remaining closing brackets).
We check if it is a valid path, if true we join remaining slices in order to get `workspaceName`.
If it is not a valid path we loop `split` backwards, joining the elements until we get a valid `workspacePath`.

What do you think about it?

GuillaumeFalourd: I don't have any problem updating the way we list workspaces.
I just think it shouldn't be done in this PR as it will impact 2 others core commands (`rit list workspace` and `rit delete workspace`) as well as `rit update workspace`.

lucasdittrichzup (dismissed):

lucasdittrichzup (approved):

lucasdittrichzup:
🚀

PR #923, opened 2021-05-12
added input flags support

fernandobelettizup:
### Description

Adding flag support to `rit set env` command.
### How to verify it
Run `rit set env --env=<env_name>` and check if it has been successfully set with `rit show env`.
### Changelog

Added flag support for rit set env command.

### DIFF
--- a/pkg/cmd/set_env.go
+++ b/pkg/cmd/set_env.go
@@ -85,8 +85,6 @@ func (s setEnvCmd) runCmd() CommandRunnerFunc {
 		env, err := s.resolveInput(cmd)
 		if err != nil {
 			return err
-		} else if env == "" {
-			return nil
 		}
 
 		if _, err := s.env.Set(env); err != nil {
--- a/pkg/cmd/set_env.go
+++ b/pkg/cmd/set_env.go
@@ -17,7 +17,9 @@
 package cmd
 
 import (
+	"errors"
 	"os"
+	"reflect"
 
 	renv "github.com/ZupIT/ritchie-cli/pkg/env"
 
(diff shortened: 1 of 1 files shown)


pkg/cmd/set_env.go
```diff
@@ -114,3 +117,44 @@ func (s setEnvCmd) runStdin() CommandRunnerFunc {
...
+}
+
+func (s *setEnvCmd) resolveFlags(cmd *cobra.Command) (string, error) {
+	env, err := cmd.Flags().GetString(setEnvFlagName)
+	if err != nil {
+		return "", err
+	} else if env == "" {
+		return "", errors.New(missingFlagText(setEnvFlagName))
+	}
+
```
henriquemoraeszup, line 158: Should you ensure this env is valid? What happens if I provide a non existent name

pkg/cmd/set_env.go
```diff
@@ -114,3 +117,44 @@ func (s setEnvCmd) runStdin() CommandRunnerFunc {
...
+	envHolder.All = append(envHolder.All, renv.Default)
+	envHolder.All = append(envHolder.All, newEnv)
+	env, err := s.List("All:", envHolder.All)
+	if err != nil {
+		return "", err
+	}
+
+	if env == newEnv {
+		env, err = s.Text("New env: ", true)
```
brunasilvazup, line 142: This `[...], true)` part guarantees values `> '' `?

pkg/cmd/set_env.go
fernandobelettizup: We do not have a `rit create env` command.
The set method, checks for an existent environment, if it does not exist it is created.

pkg/cmd/set_env.go
fernandobelettizup: Yes, if you enter a blank name, ritchie returns `Sorry, your reply was invalid: Value is required`

pkg/cmd/set_env.go
```diff
@@ -55,35 +68,25 @@ func NewSetEnvCmd(
...
 			return err
-		}
-
-		if env == newEnv {
-			env, err = s.Text("New env: ", true)
-			if err != nil {
-				return err
-			}
+		} else if env == "" {
+			return nil
```
brunasilvazup, line 89: If the value can't be nil, this logic it's necessary?

lucasdittrichzup (approved):

GuillaumeFalourd:
@fernandobelettizup : Would it be possible to add a unit test for the input flag scenario in the `set_env_test.go` file please?

GuillaumeFalourd (changes requested):
@fernandobelettizup : Would it be possible to add a unit test for the input flag scenario in the `set_env_test.go` file please?

fernandobelettizup:
@GuillaumeFalourd There are 4 tests using input flags scenario.

fernandobelettizup:
This pull request was merged via #926

PR #932, opened 2021-05-25
support flags - rit init

maurineimirandazup:
### Description

These changes are intended to improve the method of executing the `rit init` command. With them, the use of input flags will be allowed in the execution of the command.
### How to verify it
You can run the `rit init` command with the following flags:
`--sendMetrics=("yes", "no")` default: yes
`--addCommons=("yes", "no")` default: yes
`--runType=("local", "docker")` default: local

### DIFF
--- a/pkg/cmd/init.go
+++ b/pkg/cmd/init.go
@@ -262,68 +262,78 @@ func (in *initCmd) runFlags(cmd *cobra.Command) (config.Configs, error) {
 	metrics, err := cmd.Flags().GetString(metricsFlag)
 	if err != nil {
 		return config.Configs{}, err
+	} else if metrics == "" {
+		return config.Configs{}, errors.New(missingFlagText(metricsFlag))
 	}
-	common, err := cmd.Flags().GetString(commonsFlag)
+	commons, err := cmd.Flags().GetString(commonsFlag)
 	if err != nil {
 		return config.Configs{}, err
+	} else if commons == "" {
+		return config.Configs{}, errors.New(missingFlagText(commonsFlag))
 	}
 	runner, err := cmd.Flags().GetString(runnerFlag)
 	if err != nil {
 		return config.Configs{}, err
-	}
-
-	if metrics == "" {
-		return config.Configs{}, errors.New(missingFlagText(metricsFlag))
-	} else if common == "" {
-		return config.Configs{}, errors.New(missingFlagText(commonsFlag))
 	} else if runner == "" {
 		return config.Configs{}, errors.New(missingFlagText(runnerFlag))
 	}
 
-	if metrics == "no" {
-		in.metricSender.Send(metric.APIData{
-			Id:        "rit_init",
-			UserId:    "",
-			Timestamp: time.Now(),
-			Data: metric.Data{
-				MetricsAcceptance: metrics,
-			},
-		})
-	} else if metrics == "yes" {
-		if err = in.file.Write(metric.FilePath, []byte(metrics)); err != nil {
-			return config.Configs{}, err
+	switch metrics {
+	case "no":
+		{
+			in.metricSender.Send(metric.APIData{
+				Id:        "rit_init",
+				UserId:    "",
+				Timestamp: time.Now(),
+				Data: metric.Data{
+					MetricsAcceptance: metrics,
+				},
+			})
 		}
-	} else {
-		return config.Configs{}, errors.New("please provide a valid value to the flag 'sendmetrics'")
-	}
+	case "yes":
+		{
+			if err = in.file.Write(metric.FilePath, []byte(metrics)); err != nil {
+				return config.Configs{}, err
+			}
+		}
+	default:
+		{
+			return config.Configs{}, errors.New("please provide a valid value to the flag 'sendmetrics'")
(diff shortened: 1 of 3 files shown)


pkg/cmd/init.go
```diff
@@ -243,6 +229,157 @@ func (in initCmd) runStdin() CommandRunnerFunc {
...
+		return config.Configs{}, errors.New("please provide a valid value to the flag 'addCommons'")
+	}
+
+	runType := formula.DefaultRun
+	for i := range formula.RunnerTypes {
+		if RunTypes[i] == runner {
+			runType = formula.RunnerType(i)
+			break
+		}
+	}
```
brunasilvazup, line 345: Here it is not being checked if the value provided is valid as well. How about putting a null value in `runType` and checking if it's still null at the end of the for?

pkg/cmd/init.go
```diff
@@ -243,6 +229,157 @@ func (in initCmd) runStdin() CommandRunnerFunc {
...
+		}
+	case "yes":
+		{
+			if err = in.file.Write(metric.FilePath, []byte(metrics)); err != nil {
+				return config.Configs{}, err
+			}
+		}
+	default:
+		{
+			return config.Configs{}, errors.New("please provide a valid value to the flag 'sendmetrics'")
```
brunasilvazup, line 301: Here you can use the var `metricsFlag`.

pkg/cmd/init.go
```diff
@@ -57,8 +58,32 @@ var (
...
+		kind:        reflect.String,
+		defValue:    "yes",
+		description: "Download commons repository (ie: yes, no)",
+	},
+	{
+		name:        runnerFlag,
+		kind:        reflect.String,
+		defValue:    "local",
+		description: "Set runner type (ie: local, docker)",
+	},
```
brunasilvazup, line 84: Hi @maurineimirandazup,
I don't know if these variables should have default values ​​set, because by some error the user may have a choice that he didn't make (download new repositories, accept having his data collected).
Maybe @GuillaumeFalourd can contribute, but I think it's safer for the user not to have the answers by default.

And you can use the kind `reflect.Bool` if you want restrict the values, but I dont know how to `true or false` can be accept to UX.

pkg/cmd/init.go
```diff
@@ -243,6 +229,157 @@ func (in initCmd) runStdin() CommandRunnerFunc {
...
+		return config.Configs{}, errors.New(missingFlagText(commonsFlag))
+	}
+	runner, err := cmd.Flags().GetString(runnerFlag)
+	if err != nil {
+		return config.Configs{}, err
+	} else if runner == "" {
+		return config.Configs{}, errors.New(missingFlagText(runnerFlag))
+	}
+
+	switch metrics {
```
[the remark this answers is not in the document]
maurineimirandazup: This was the initial proposal, but in stdin and in the prompt, these logics were already repeated. In an internal discussion, it was decided to open an ISSUE to refactor all this part into the next version.

pkg/cmd/init.go
maurineimirandazup: The proposal to use reflect.Bool was indeed interesting, but for that we would have to make changes to the flags lib, so as discussed, I used another approach that comes close to this idea.

pkg/cmd/init.go
maurineimirandazup: Done.

pkg/cmd/init.go
```diff
@@ -243,6 +229,157 @@ func (in initCmd) runStdin() CommandRunnerFunc {
...
+			repo.Version = formula.RepoVersion(tag.Name)
+			if err := in.repo.Add(repo); err != nil {
+				s.Error(ErrInitCommonsRepo)
+				fmt.Println(addRepoMsg)
+				return config.Configs{}, err
+			}
+			in.commonsSuccess(s)
+		}
+	default:
+		return config.Configs{}, errors.New("please provide a valid value to the flag 'addCommons'")
```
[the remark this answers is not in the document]
maurineimirandazup: Done.

pkg/cmd/init.go
maurineimirandazup: Done.

pkg/cmd/init.go
```diff
@@ -403,3 +544,21 @@ func (in initCmd) initSuccess() {
 	success := i18n.T("init.successful")
 	prompt.Success(success)
 }
+
+func flagToBool(f string, fn string) (bool, error) {
+	boolOpts := map[string]bool{
```
GuillaumeFalourd, line 549: It could be interesting to use the map variable from this class instead: https://github.com/ZupIT/ritchie-cli/blob/af0a183d47cbf2f814abb23c38407f4a77223621/pkg/prompt/bool.go#L24 to avoid repeating the same code.

GuillaumeFalourd (changes requested):

GuillaumeFalourd (approved):
Just fix the linter. The rest looks good to me, well done!👍🏼

GuillaumeFalourd (approved):
👍🏼

brunasilvazup (approved):

PR #938, opened 2021-05-26
POC - Github Action Workflow to test RIT binaries for each supported OS

GuillaumeFalourd:
### Description

- This contribution is a Proof Of Concept (POC) showing how to perform funcional tests using **rit** binaries for each OS, for the `repo` core commands. Those kind of tests could improve the project quality and future pipeline performances by testing only what matters on each contribution, instead of testing everything every time.

- This workflow will trigger every time one of the paths informed files is updated, by a `PULL REQUEST` event or a `PUSH` event.

- The workflow executes in **parallel** jobs on `windows`, `ubuntu` and `macOS` where each job will:
    - use the make command to generate rit binary for the OS.
    - execute synchronously the following steps: `rit list repo`, `rit add repo`, `rit update repo`, `rit delete repo`.
    - after each step, a check run is performed using `diff` command to assert the output is as expected.

- Expected outputs need to be added to the `/testdata/gha_workflows` directory, currently on the `.txt` format

#### Possible enhancements

- It is not possible to test **expected** error that way without breaking the job execution (which result in a workflow error). So it could be interesting to improve the way those functional tests are performed (through a script or a formula?) to be more assertive.

### How to verify it

- You can check the workflow execution on [this fork](https://github.com/GuillaumeFalourd/ritchie-cli/actions/workflows/test-repo-commands.yml)

Here are some samples:

workflow

job

### Changelog

- Adding Github Action workflow performing funcional tests using **rit** binaries for each OS, for the `repo` core commands.


lucasdittrichzup (approved):
🚀

PR #956, opened 2021-06-08
Github Action workflow to perform functional tests for `CREDENTIAL` core commands using RIT binaries

GuillaumeFalourd:
### Description

- Funcional tests using **rit** binaries for each OS, for the `credential` core commands. Those kind of tests could improve the project quality and future pipeline performances by testing only what matters on each contribution, instead of testing everything every time.

- This workflow will trigger every time one of the paths informed files is updated, by a `PULL REQUEST` event or a `PUSH` event.

- The workflow executes in **parallel** jobs on `windows`, `ubuntu` and `macOS` where each job will:
    - use the make command to generate rit binary for the OS.
    - execute synchronously the following steps: `rit list credential`, `rit set credential`, `rit delete credential`.
    - after each step, a check run is performed using `diff` command to assert the output is as expected.

- Expected outputs need to be added to the `/testdata/gha_workflows` directory, currently on the `.txt` format

#### Possible enhancements

- It is not possible to test **expected** error that way without breaking the job execution (which result in a workflow error). So it could be interesting to improve the way those functional tests are performed (through a script or a formula?) to be more assertive.

### How to verify it

- You can check the workflow execution on [this fork](https://github.com/GuillaumeFalourd/ritchie-cli/actions/workflows/test-credential-commands.yml)

Here are some samples:

Screen Shot 2021-06-08 at 13 21 44

Screen Shot 2021-06-08 at 13 21 53

[Reference](https://github.com/GuillaumeFalourd/ritchie-cli/runs/2775134700)

### Changelog

- Adding Github Action workflow performing funcional tests using **rit** binaries for each OS, for the `credential` core commands.


.github/workflows/test-credential-commands.yml
```diff
@@ -0,0 +1,125 @@
...
+jobs:
+  ubuntu:
+      runs-on: ubuntu-latest
+      steps:
+        - uses: actions/checkout@v2.3.4
+        - name: Create binary from branch
+          run: |
+            cd $GITHUB_WORKSPACE
+            make build-linux && sudo mv ./dist/linux/rit /usr/local/bin
+            echo '{"addCommons":false, "sendMetrics":true, "runType":"local"}' | rit init --stdin
```
lucasdittrichzup, line 56: sending metrics through the pipelines won't distort them? Since we're talking about tests, sending these values can make extraction difficult later, no?

lucasdittrichzup (approved):

.github/workflows/test-credential-commands.yml
GuillaumeFalourd: Done @lucasdittrichzup 👍🏼  

As you will see, I also updated all workflows with the new rit init command using input flags:
- https://github.com/GuillaumeFalourd/ritchie-cli/actions/runs/936031856
- https://github.com/GuillaumeFalourd/ritchie-cli/actions/runs/936016363
- https://github.com/GuillaumeFalourd/ritchie-cli/actions/runs/936016178

lucasdittrichzup (approved):

